PR #7, opened 2016-10-09
Backtracking line search with interpolation

cortner:
This PR adds a second backtracking line search, `interpbacktrack_linesearch!` which, instead of an a *= rho update performs an interpolation step; this is highly efficient for some problems; see QCG and QLBFGS in the following figure (from my own research); 
optim

It also outperforms HZ and MT line-search for some more standard model problems.  In general, it is no worse than standard backtracking, which I left only for the sake of compatibility. I would actually recommend to remove standard backtracking altogether and replace it with `interpbacktrack`.

### DIFF
--- a/src/backtracking_linesearch.jl
+++ b/src/backtracking_linesearch.jl
@@ -7,6 +7,8 @@ f(0), f'(0), f(α) has a minimiser α' in the open interval (0, α). More strong
 there exists a factor ρ = ρ(c₁) such that α' ≦ ρ α. This makes
 the `interpbacktrack_linesearch!` a backtracking type linesearch.
 
+This is a modification of the algorithm described in Nocedal Wright (2nd ed), Sec. 3.5.
+
 **Default Parameters**
 
 * `c1 = 0.2` : Armijo condition
--- a/src/backtracking_linesearch.jl
+++ b/src/backtracking_linesearch.jl
@@ -1,3 +1,32 @@
+
+# wrapping backtracking_linesearch! but calling it with interp=true
+interpbacktrack_linesearch!{T}(df,
+                               x::Vector{T},
+                               s::Vector,
+                               x_scratch::Vector,
+                               gr_scratch::Vector,
+                               lsr::LineSearchResults,
+                               alpha::Real = 1.0,
+                               mayterminate::Bool = false,
+                               c1::Real = 0.2,
+                               c2::Real = 0.9,
+                               rho=0.9,
+                               iterations::Integer = 1_000) =
+   backtracking_linesearch!(df,
+                             x,
+                             s,
+                             x_scratch,
+                             gr_scratch,
+                             lsr,
+                             alpha,
+                             mayterminate,
+                             c1,
+                             c2,
+                             rho,
+                             iterations,
+                             true)
+
+
 function backtracking_linesearch!{T}(df,
                                      x::Vector{T},
                                      s::Vector,
@@ -9,7 +38,21 @@ function backtracking_linesearch!{T}(df,
                                      c1::Real = 1e-4,
                                      c2::Real = 0.9,
                                      rho::Real = 0.9,
-                                     iterations::Integer = 1_000)
+                                     iterations::Integer = 1_000,
+                                     interp::Bool = false)
+
+    # Check the input is valid, and modify otherwise
+    if interp   # this means we are coming from interpbacktrack_linesearch!
+       backtrack_condition = 1.0 - 1.0/(2*rho) # want guaranteed backtrack factor
+       if c1 >= backtrack_condition
+           warning("""The Armijo constant c1 is too large; replacing it with
+                      $(backtrack_condition)""")
+           c1 = backtrack_condition
+       end
+       if rho <= 0.25
+           warn("rho <= 0.25; revert to standard backtracking")
+       end
+    end
 
     # Count the total number of iterations
     iteration = 0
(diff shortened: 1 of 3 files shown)


src/backtracking_linesearch.jl
```diff
@@ -46,8 +85,21 @@ function backtracking_linesearch!{T}(df,
...
+        else
+           # backtracking via interpolation:
+           # This interpolates the available data
+           #    f(0), f'(0), f(α)
+           # with a quadractic which is then minimised; this comes with a
+           # guaranteed backtracking factor 0.5 * (1-c1)^{-1} which is < 1
+           # provided that c1 < 1/2; the backtrack_condition at the beginning
+           # of the function guarantees at least a backtracking factor rho.
+           alpha1 = - (gxp * alpha) / ( 2.0 * ((f_x_scratch - f_x)/alpha - gxp) )
+           alpha = max(alpha1, alpha * min(0.25, rho))  # avoid miniscule steps
```
anriseth: Is there a heuristic for choosing 0.25 over other numbers?

anriseth: I think a `rhomin` parameter defaulted to 0.25 can be useful.

src/backtracking_linesearch.jl
```diff
@@ -9,7 +38,21 @@ function backtracking_linesearch!{T}(df,
...
+    # Check the input is valid, and modify otherwise
+    if interp   # this means we are coming from interpbacktrack_linesearch!
+       backtrack_condition = 1.0 - 1.0/(2*rho) # want guaranteed backtrack factor
+       if c1 >= backtrack_condition
+           warning("""The Armijo constant c1 is too large; replacing it with
+                      $(backtrack_condition)""")
+           c1 = backtrack_condition
+       end
+       if rho <= 0.25
+           warn("rho <= 0.25; revert to standard backtracking")
```
anriseth: It seems to me that we only revert back to standard backtracking if
`alpha1` < `alpha * min(0.25, rho)`. Is that correct, or have I misunderstood?

anriseth:
Thanks! I have found the current backtracking algorithm to be poor-performing, so this is great.
Is there a source describing the approach that we can reference, in case people want to better understand the idea behind the interpolation step?

An alternative to removing backtracking, can be to make the `interp` flag true by default?
I think we should make a `NEWS.md` similar to `Optim` to keep track of these changes as well.

cortner:
The problem with just setting interp=true is it then gets awkward to pass backtracking as an option.

Sent from my iPhone

On 9 Oct 2016, at 19:12, Asbj?rn Nilsen Riseth <notifications@github.com<mailto:notifications@github.com>> wrote:


Thanks! I have found the current backtracking algorithm to be poor-performing, so this is great.
Is there a source describing the approach that we can reference, in case people want to better understand the idea behind the interpolation step?

An alternative to removing backtracking, can be to make the interp flag true by default?
I think we should make a NEWS.md similar to Optim to keep track of these changes as well.

-

src/backtracking_linesearch.jl
cortner: `interpbacktrack` ensures that at each step `alpha` is decreased by at least a factor `rho`. So if `rho < 0.25` (or, `rho <= mindecrease`) then it will be standard backtracking with factor `rho`.

cortner:
Funnily enough I've not seen this anywhere, hence I mentioned it in an appendix of a paper where I used it. But this certainly doesn't mean others haven't done it -it is an extremely simple and natural idea. Maybe if you run into Nick Gould at some point ask him?

src/backtracking_linesearch.jl
cortner: `mindecfact`

src/backtracking_linesearch.jl
```diff
@@ -9,7 +54,24 @@ function backtracking_linesearch!{T}(df,
...
-                                     iterations::Integer = 1_000)
+                                     iterations::Integer = 1_000,
+                                     interp::Bool = false,
+                                     mindecfact=0.25)
+
+    # Check the input is valid, and modify otherwise
+    if interp   # this means we are coming from interpbacktrack_linesearch!
+       backtrack_condition = 1.0 - 1.0/(2*rho) # want guaranteed backtrack factor
+       if c1 >= backtrack_condition
+           warning("""The Armijo constant c1 is too large; replacing it with
```
anriseth: Should this be `warn`? I couldn't find a `warning` function in `Base` on Julia 0.5.

src/backtracking_linesearch.jl
```diff
@@ -21,13 +83,9 @@ function backtracking_linesearch!{T}(df,
...
     n = length(x)
 
-    # Store f(x) in f_x
-    f_x = df.fg!(x, gr_scratch)
-    f_calls += 1
-    g_calls += 1
-
-    # Store angle between search direction and gradient
-    gxp = vecdot(gr_scratch, s)
+    # read f_x and slope from LineSearchResults
```
anriseth: This is useful. It seems like `Optim` is always passing in an up to date `LineSearchResults`. 
@KristofferC: Will taking `f_x` and `gxp` from `lsr` break `NLOpt`s use of line searches, or is this fine?

src/backtracking_linesearch.jl
cortner: right - thank you for catching this.

src/backtracking_linesearch.jl
cortner: Btw, I think this needs to be cleaned up in other linesearches as well. According to @pkofod only HZ uses this so far.

anriseth:
I had a look in Nocedal and Wright - Numerical Optimization. 
This algorithm is described as an enhancement to standard backtracking, found in Chapter 3, the section on Interpolation (page 56 in my edition).

They go one step further: 
In the first iteration they use quadratic interpolation, and then a cubic interpolation with the two previous alpha-values, i.e. f(0), f'(0), f(α_{k}), f(α_{k-1})

cortner:
great - I didn’t realise they discussed this.

I specifically don’t want to use the cubic though. I’ve found this quadratic interpolation to be extremely robust. But if we can introduce the linesearch options, then we can just let the user choose which interpolation to use.

cortner:
I added the reference now. It would be good to add the cubic interpolation as well. Hopefully we can do that in a separate pull request after moving to LineSearchOptions.

anriseth:
Great, I'll merge later today.

PR #14, opened 2016-10-11
Use lsr.value and lsr.slope

### LINKED ISSUE #11: Get objective value and slope from `LineSearchResults`
Ref: https://github.com/anriseth/LineSearches.jl/pull/7#discussion_r82532245

Some of the linesearch algorithms calls the objective unnecessarily when calculating the value at the initial `x`.

anriseth:
In `mt_linesearch!` and `interpolating_linesearch!`, the objective value and slope is calculated for the initial `x`, causing an extra function and gradient call.
Optim already stores this information in the LineSearchResults object passed to the algorithms.

We now re-use the already calculated objective value and slope at the initial `x`.

Closes #11


anriseth:
@KristofferC, in case this is useful for `NLSolve`

anriseth:
Hmm. Is it a bad idea to make the tests in LineSearches require Optim?

Issue #9, opened 2016-10-09
LineSearchOptions


cortner:
At the moment all linesearchoptions (e.g. C1, C2) are passed via arguments. This makes it a bit of a pain for the user to set those arguments given how the line search functions are passed to Optim. I suggest to either move to a `LineSearchOptions` type or to keyword arguments.

anriseth:
I agree. Pros and cons of keyword arguments vs a `LineSearchOptions` type?

Keywords is the easiest change to make. Will `LineSearchOptions` give us more power + flexibility than plain kwargs, or potentially be less messy?

cortner:
Another option would be to make a linesearch a type that stores the options, the function `linesearch!` would then dispatch to the correct linesearch. 

I actually like this by far the best.

cortner:
```
optimise!(f, g!, method=LBFGS(), 
                         linesearch=BackTracking(interp=:cubic, rho=0.9))
```

cortner:
@pkofod what do you think about this; I think it will make the Optim codes much cleaner.

anriseth:
Sorry, I think I didn't register the two lasts posts when you wrote them.
Something along this line with linesearch and alphaguess types/routines sound like a good idea for flexibility.

pkofod:
Yes, I think we need something like this. I'll think, and return.

cortner:
I am still offering to help, but (like you) I just had another child born so any work I can do on this will be sporadic at best for the next few months

pkofod:
> I am still offering to help, but (like you) I just had another child born so any work I can do on this will be sporadic at best for the next few months

Congrats! Hope you're all doing great. I think we're pretty close to being at a point where we should have a stab at this, right @anriseth ? I say we go the dispatch route. It will be the easiest way for the user to flexibly specify how to perform the line search (just like without `Optimizer` types in Optim).

anriseth:
Yes, the dispatch route is great.
Would it make sense to start this after implementing the *Function changes in NLBase? Or would you say we can start this in parallel? (I'm thinking in terms of merging issues on top of existing PRs)

pkofod:
> Yes, the dispatch route is great.
> Would it make sense to start this after implementing the *Function changes in NLBase? Or would you say we can start this in parallel? (I'm thinking in terms of merging issues on top of existing PRs)

[...]

PR #20, opened 2016-11-14
RFC: Implement cubic backtracking

### LINKED ISSUE #13: Implement backtracking with cubic interpolation
We should redo the backtracking algorithm, and let the user choose between quadratic and cubic interpolation.

The "standard" case with no interpolation does not work very well, so I say we scrap that.

Follow "Numerical Optimization" by Nocedal and Wright, page 56, and/or
"Numerical Methods for Unconstrained Optimization and Nonlinear Equations" by Dennis and Schnabel, page 325.

anriseth:
I've updated the backtracking algorithm to use either quadratic or cubic interpolation as described in Nocedal and Wright. Parameters are taken from there and from "Numerical Methods for Unconstrained Optimization and Nonlinear Equations" by Dennis and Schnabel, page 325.

This might break people's code, as I've updated the parameter list list for the algorithm (sorting out #9 will make changes like this much smoother)

Closes #13

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# LineSearches v0.1.3 release notes
+* The backtracking algorithm now performs quadratic or cubic interpolations
+[#20](https://github.com/anriseth/LineSearches.jl/pull/20)
+
 # LineSearches v0.1.1 release notes
 * Hagerzhang and Backtrack now throw LineSearchExceptions
 [#19](https://github.com/anriseth/LineSearches.jl/pull/19)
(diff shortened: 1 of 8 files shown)


anriseth:
@cortner FYI, your `interpbacktrack!` is now `bt2!`, but I have also changed the default values for c1 and c2 to the values they use in PETSc / "Numerical Methods for Unconstrained Optimization and Nonlinear Equations" by Dennis and Schnabel, page 325.

cortner:
I know that a lot of texts recommend c1 = 1e-4, but it have never encountered a case in practise where this is a good choice. I have always found that values between 0.1 and0.25 are best when you use quadratic or cubic interpolation.

PR #26, opened 2017-01-10
Add basic linesearch

### LINKED ISSUE #25: Create basic linesearch with predefined step-length
Can be useful for example with Newton's method in the quadratic regime.

Similar to 
https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/SNES/SNESLINESEARCHBASIC.html

This should also help #24

anriseth:
Fixes #25


anriseth:
The tests needs to fix https://github.com/JuliaOpt/Optim.jl/issues/339 in order to work.

PR #27, opened 2017-01-20
Add Julia 0.6 to Travis

anriseth:


pkofod:
Good idea. If feature freeze is close, we might as well start fixing stuff now rather than later :)

pkofod:
Any reason why you're bumping to v0.5? I intended to do so in optim as well, just curious.

anriseth:
> Any reason why you're bumping to v0.5? I intended to do so in optim as well, just curious.

It's because you were gonna do it in Optim, and the next major here will probably follow Optim's next major

PR #33, opened 2017-02-09
Add fcount, gcount tests

anriseth:
Test that all the linesearches algorithms send back the correct fcounts to Optim result.


pkofod:
https://github.com/JuliaLang/METADATA.jl/pull/7912

PR #37, opened 2017-03-15
Use NLSolversBase

anriseth:
Needs https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/1
For the tests to work, we also need to remove the objective_types from Optim


anriseth:
Two things to think about:
- Should we remove the `gr_new` gradient storage from linesearches? This isn't really needed anymore, as the values can be fetched from `df.g` since all calls use `value_grad!` and `gradient!`.
- NLSolversBase is already updating f_calls and g_calls for us now. Do we want to keep counting them within linesearches as well, and why?

pkofod:
> NLSolversBase is already updating f_calls and g_calls for us now. Do we want to keep counting them within linesearches as well, and why?

I don't think we do. If someone wants to know how many f and g calls were made during line search, I think it would be better to compare a before and after. That way you don't risk over or under counting them by mistake (for example, Nelder Mead had some f calls there didn't increment f_calls).

anriseth:
> I don't think we do. 

Great. Then I'll remove the f_calls and g_calls return values in this PR as well.

PR #38, opened 2017-04-06
Adhere to new storage/x argument order, and bump NLSolversBase versio…

pkofod:
…n to 2.0.0 and up.

This is all there is to it I guess? I mean, we kind of created NLSolversBase such that changes like the one we are facing wouldn't really matter in this code.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,2 @@
 julia 0.5
-NLSolversBase 1.0 2.0
+NLSolversBase 2.0
--- a/test/alphacalc.jl
+++ b/test/alphacalc.jl
@@ -3,8 +3,8 @@ let
 
     lsalphas = [1.0, 0.5,0.5,0.49995,0.5,0.5,0.5]
 
-    f(x) = vecdot(x,x)
-    function g!(x, out)
+    f(x) = vecdot(x, x)
+    function g!(out, x)
         out[:] = 2x
     end
 
--- a/test/counter.jl
+++ b/test/counter.jl
@@ -31,9 +31,9 @@ let
         fcounter()
         prob.f(x)
     end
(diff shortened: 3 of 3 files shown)


pkofod:
Hm, we don't have AV ci here?... We should

pkofod:
@anriseth do you agree that nothing substantial needs to be changed besides the tests? After we make the change in Optim we should enable the Optim tests again.

anriseth:
Thanks. Yeah with the value and grad api in Base, the tests are the only thing that needs changing.
But we will need to make sure all packages that rely on LineSearches and Base have the correct upper limits. I think that's only Optim for now?

pkofod:
I think nlsolve has the correct upper bound. Wasn't that what we went through in metadata the other week?

anriseth:
Yep, they're still on pre-v1.0.

I'll merge this now

On Thu, 6 Apr 2017, 18:00 Patrick Kofod Mogensen, <notifications@github.com>
wrote:

I think nlsolve has the correct upper bound. Wasn't that what we went
through in metadata the other week?

—
You are receiving this because you were mentioned.


Reply to this email directly, view it on GitHub
<https://github.com/JuliaNLSolvers/LineSearches.jl/pull/38#issuecomment-292238343>,
or mute the thread

anriseth:
Hmm, I don't understand why the Travis push test fails, whilst the Travis pr test works?

pkofod:
Oh, that doesn't matter, it's just because it's targeting a different master than current master, and I removed the optim dependency in tests after I created this pr, because we're not using them yet.

pkofod:
just to prove that it was okay: https://travis-ci.org/JuliaNLSolvers/LineSearches.jl :)

PR #42, opened 2017-04-07
Fix appveyor

anriseth:
I think this should fix the appveyor error


pkofod:
Good old white space!

PR #43, opened 2017-05-08
Types

cortner:
I suggest to not merge this yet. This is just a proof-of-concept for `backtracking!` only. Please have a quick look - if you are happy then I'll try to do the same for the remaining line searches.

Notes: 
 * This PR will give - for the user - the functionality of using types instead of functions for LS, but internally very little has changed. Both of you (@pkofod and @anriseth)  are probably better placed to carry out the internal changes. (though time permitting I'm obviously happy to do this too)
 * While I was at it, I also made a small addition to `backtracking!` adding a `maxstep` parameter; this requires that `alpha * vecnorm(s, inf) <= maxstep`. (the choice of norm is a little restrictive, but I'll do another PR later when this is done to generalise to arbitrary step-length constraints).

Please, somebody double-check that I've deprecated `bt2!` and `bt3!` correctly.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,3 @@
 julia 0.5
 NLSolversBase 2.0
+Parameters
--- a/src/LineSearches.jl
+++ b/src/LineSearches.jl
@@ -11,9 +11,8 @@ export LineSearchResults, LineSearchException
 
 export clear!, alphatry, alphainit
 
-export hagerzhang!, backtracking!, strongwolfe!,
-    morethuente!, bt2!, bt3!, basic!,
-    BackTracking, HagerZhang, Static, MoreThuente, StrongWolfe
+export BackTracking, HagerZhang, Static, MoreThuente, StrongWolfe
+    
 
 include("types.jl")
 include("api.jl")
--- a/src/LineSearches.jl
+++ b/src/LineSearches.jl
@@ -13,7 +13,7 @@ export clear!, alphatry, alphainit
 
 export hagerzhang!, backtracking!, strongwolfe!,
     morethuente!, bt2!, bt3!, basic!,
-    BackTracking, HagerZhang, Static
+    BackTracking, HagerZhang, Static, MoreThuente, StrongWolfe
 
 include("types.jl")
 include("api.jl")
--- a/src/LineSearches.jl
+++ b/src/LineSearches.jl
@@ -13,7 +13,7 @@ export clear!, alphatry, alphainit
 
 export hagerzhang!, backtracking!, strongwolfe!,
     morethuente!, bt2!, bt3!, basic!,
-    BackTracking
+    BackTracking, HagerZhang, Static
 
 include("types.jl")
 include("api.jl")
--- a/src/LineSearches.jl
+++ b/src/LineSearches.jl
@@ -2,6 +2,8 @@ isdefined(Base, :__precompile__) && __precompile__()
 
 module LineSearches
 
+using Parameters
+
 import NLSolversBase
 import Base.clear!
 
@@ -10,7 +12,8 @@ export LineSearchResults, LineSearchException
 export clear!, alphatry, alphainit
 
 export hagerzhang!, backtracking!, strongwolfe!,
-    morethuente!, bt2!, bt3!, basic!
+    morethuente!, bt2!, bt3!, basic!,
+    BackTracking
 
 include("types.jl")
 include("api.jl")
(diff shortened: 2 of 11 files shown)


anriseth:
This looks good to me. Seems like it is quite straightforward to introduce types :)

I think that, while we're at it - we could evaluate whether the `gr_scratch` parameter is necessary. `value_gradient!` and `gradient!` makes it obsolote, as far as I can remember (was looking at this a few months ago)

cortner:
Ok - will try to finish this over the course if the next few days

cortner:
This should in principle complete the PR. Some more notes: 

I've called the basic line search `Static` rather than `Basic`, this seems more descriptive to me. 

At this point it would be good to decide whether the `call` trick is actually sufficient, or whether - in a second step - it would be better to replace it, i.e., 
```
(ls::LineSearch)(args...)
```
vs
```
linesearch(ls, args...) 
```

Either way, the next step - maybe still within this PR - might be to deprecate all of the line search functions `morethuente!, hagerzhang!`, etc and replace them with *private* functions `_morethuente!`, etc.  (On that note: how do I create a custom deprecation warning?)

pkofod:
> I've called the basic line search Static rather than Basic, this seems more descriptive to me.

I agree; basic could just as well be backtracking to some people
 
> At this point it would be good to decide whether the call trick is actually sufficient, or whether - in a second step - it would be better to replace it, i.e.,

> (ls::LineSearch)(args...)
> 
> vs
> 
> linesearch(ls, args...) 

why would the second be preferred?

> Either way, the next step - maybe still within this PR - might be to deprecate all of the line search functions morethuente!, hagerzhang!, etc and replace them with private functions _morethuente!, etc. (On that note: how do I create a custom deprecation warning?)

yes, this needs to be done, and I think the "privatization" is a good idea. Let me get back to you on the "custom" warning unless someone gets to it before me (I need to go)

pkofod:
regarding the custom dep, you can do something like
```jl
const has_deprecated_xyz = Ref(false)

function xyz(args)
        if !has_deprecated_xyz![]
            warn("xyz is deprecated, please ... descriptive message here ...")
            has_deprecated_xyz![] = true
        end
    ... code here ...
    return ...
end
```

cortner:
ok so pretty much hand-code it, that will be fine.

cortner:
`(ls::LineSearch)(args...)` vs `linesearch(ls, args...)`

> why would the second be preferred? 

The second seems the more natural construction in the dispatch context, but it really doesn't have to be. I am perfectly happy with making linesearch types callable.

cortner:
> `const has_deprecated_xyz = Ref(false)`

why `Ref`?

cortner:
see the [PR #411 in Optim](https://github.com/JuliaNLSolvers/Optim.jl/pull/411) for the corresponding changes in Optim.

anriseth:
(ls::LineSearch)(args...) vs linesearch(ls, args...)
> why would the second be preferred?

The dispatch approach would align with Optim's `optimize` approach. So maybe a preference due to consistency?
I haven't tried using callable types much, so I don't know how they work and whether there are any pros and cons vs dispatch. It would be interesting to hear if anyone knows:)

anriseth:
Ready to merge?

> `const has_deprecated_xyz = Ref(false)`
> why Ref?

I'm curious as well

cortner:
hm - when I was writing it, it started to make sense, I'll need to look at the code again.

cortner:
I remember now - this is actually the only way to do it - at least as I've implemented the deprecation in the end. If you are more comfortable with macros than I am then you could probably rewrite these deprecations much more elegantly though.

anriseth:
I see. No I'm not a big macro genius - just hadn't seen Ref before.

Is this ready to merge?

cortner:
from my end, yes it is ready

pkofod:
hmm, can't really remember the technical details... I got it from @KristofferC when I used it in Optim :)

pkofod:
The function approach would be fine by me. It's currently called `perform_linesearch` in Optim, but doesn't have to have the `perform` included for all I care.

anriseth:
Great, thanks!

pkofod:
I think we should actually define the function (whatever name we choose) in `NLSolversBase`. This would mean that another linesearch package could just `using NLSolversBase`, adhere to the same syntax, and then that would work out of the box in Optim.

cortner:
Does it also become important then to define an abstract supertype `AbstractLineSearch`? I don't really see the necessity here, but maybe I am missing something?

PR #48, opened 2017-05-20
Remove gradient storage

### LINKED ISSUE #47: Remove gradient storage array from linesearches?
All the linesearches take an argument `g::Array` or similar, which was previously used to store the gradient. With the use of NLSolversBase and `*gradient!`, this seems unnecessary?

I'll remove it, if nobody has an objection @cortner @pkofod

anriseth:
Fixes #47

I also updated the README and the tests (can't enable them until Optim 1.0 is tagged)


src/hagerzhang.jl
```diff
@@ -84,7 +82,6 @@ function _hagerzhang!{T}(df,
                         x::Array{T},
                         s::Array,
                         xtmp::Array,
-                        g::Array,
```
pkofod: If you keep a signature with a warning and `g_scratch`, and make sure that `copy!(g_scratch, d.g` at the end of this method, like

```
function _hagerzhang!(df, x, s, xtmp, g, lsr, c, mayterminate, ...)
    warn("you no longer have to provide a "g" input")
    _hagerzhang!(df, x, s, xtmp, lsr, c, mayterminate, ...)
    copy!(g, d.g)
end
```
then this doesn't have to be a breaking change... don't know if it's worth the trouble though.

anriseth: Good idea! I'll try to do this.

anriseth:
This can be merged, if nobody objects.

Then we can tag a new minor

pkofod:
I'll merge after the tests run (can't be sure the conflict resolution worked flawlessly)

PR #54, opened 2017-06-20
Replace @simd loops

anriseth:


pkofod:
Just for fun, did you try to benchmark it? I don't suspect a difference (at least not for the worse) just curious.

anriseth:
I haven't benchmarked it. Blind faith ;)

pkofod:
I like it!

pkofod:
On a serious note I do find that the simd version is hard to predict perf on. In NelderMead it slowed some parts down. Guess I just don't know enough about simd vectorization.

PR #55, opened 2017-06-25
Accept multidimensional arrays as input

antoine-levitt:
Simple vector->array. I also typed everything to T for consistency.


anriseth:
Thank you for improving all the obsolete overtyping @antoine-levitt :)
Is there a reason for why we are stopping at `Array`?

antoine-levitt:
Not sure, that depends on whether concrete types in structs are important or not. An alternative is to force static typing by using parametric types, but there must be a downside to doing that and just parametrizing the type of every field in a struct, otherwise why would anybody ever do anything else?

anriseth:
I was thinking in terms of using, for example, AbstractArray? Would that
make a difference for performance? I'm happy with Array, just wondering.

antoine-levitt:
The API is wrong in any case, because linesearch should only see, well, line searches and the API should be rewritten, so this conversation is best had at the Optim level (see https://github.com/JuliaNLSolvers/Optim.jl/issues/399)

PR #59, opened 2017-10-10
Remove Optim related tests.

pkofod:
So... this leaves LineSearches.jl in a kind of sorry state. We probably need to test this package better. I'm moving these tests to another package later.

I should mention that  this has been discussed on gitter.


pkofod:
cc: @anriseth

anriseth:
Ways to replace Optim tests
- [ ] Cubic backtracking: Extract the info sent by BFGS on Himmelblau that hits https://github.com/JuliaNLSolvers/LineSearches.jl/blob/master/src/backtracking.jl#L85
- [ ] Counter: Find a situation where all the (non-trivial) linesearches require at least two evals (e.g. Himmelblau above?)
- [ ] Optim usage: This should be covered by the "other package", and not in LineSearches. 

I can deal with the first and second one.

test/optim.jl
```diff
@@ -1,23 +0,0 @@
...
-
-    for ls in lstypes
-        opts = Optim.Options(allow_f_increases = true)
-
-        debug_printing && println("\nTesting $(string(ls))")
-        counter = 0
-        for (name, prob) in Optim.UnconstrainedProblems.examples
-            if prob.isdifferentiable
-                counter +=1
-                if counter != 4
```
anriseth: Not sure why this counter test was here (I added it in #50). Confused.

pkofod:
> I can deal with the first and second one.

You should be able to push to this branch, right?

anriseth:
> You should be able to push to this branch, right?
I'll try :p

anriseth:
To do what I said, I still need the Himmelblau test function. For now, we can just define it again in the test system. I think the test functions should belong to a separate package from Optim, however.

pkofod:
> For now, we can just define it again in the test system. I think the test functions should belong to a separate package from Optim, however.

They should definitely.

pkofod:
Right, so this is good to go right?

pkofod:
> Edit: I actually had such a package finished, but I broke my computer and the code was only local. I'll remake it, but just copy it over for now.

Oh, I actually did push it before I broke my computer https://github.com/JuliaNLSolvers/OptimTestProblems.jl lol, it's just not tagged.

anriseth:
> Right, so this is good to go right?

Sorry, I got a bit carried away and added more tests.

Good to go now (assuming tests don't fail)

PR #64, opened 2017-10-10
Enable api tests + other

anriseth:
I forgot to enable the API tests in #59 ...
Also added another test.

Is there a way to check coverage without pushing to remote?


pkofod:
https://github.com/JuliaCI/Coverage.jl

PR #66, opened 2017-10-11
Use OptimTestProblems

pkofod:


anriseth:
Ah, you were a few minutes ahead of me :+1: 

The `backtracking.jl` test is identical to the Himmelblau-test for `BackTracking(order=3)` in `alphacalc.jl` now, so we can delete `backtracking.jl`.

pkofod:
Oh, I'm sorry! We should coordinate better :)

anriseth:
In the name of coordination: Would you like to tag `3.0.1` after this PR? ;)

PR #67, opened 2017-10-26
Rename basic.jl to static.jl

anriseth:
Are we using `NEWS.md`? Or are the release notes for each version enough?


pkofod:
> Are we using NEWS.md? Or are the release notes for each version enough?

I often forget - in Optim as well...but we can if you want.

PR #68, opened 2017-10-30
Allow scaled versions of Static

anriseth:
For the Static linesearch, I have use cases where it is useful to have a step size that is static, relative to the direction norm.

### DIFF
--- a/src/static.jl
+++ b/src/static.jl
@@ -5,13 +5,27 @@
 """
 `Static`: defines a static linesearch, i.e. with fixed step-size. E.g., initialise
 with `Static(alpha = 0.3141)` for fixed step-size 0.3141. Default is 1.0.
+
+You can also make this independent of the size of the step `s`, by using
+`Static(scaled = true)`.
+This will then use a steps-size alpha ← min(ls.alpha,||s||_2) / ||s||_2
 """
 @with_kw struct Static{T}
     alpha::T = 1.0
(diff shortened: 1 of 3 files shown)


pkofod:
@anriseth is it really necessary to do the `s` check? It makes it kind of hard to remove `s` as an input :)

anriseth:
I'm not sure what you mean by the `s` check? 

In any case, I realize this causes some issues with the univariatization. Here's one approach that may let us still keep the scaled approach: 

- Redo Static to take step length provided by `ls(..., alpha,...)`, and remove `Static.alpha`.
- Introduce `InitialStatic.scaled` instead, and let that provide the initial `alpha = min(is.alpha, ns) / ns`

pkofod:
> Introduce InitialStatic.scaled instead, and let that provide the initial alpha = min(is.alpha, ns) / ns

Uhm... lol.. seems like it is *already* in `InitialStatic`, so basically this is done twice...

https://github.com/JuliaNLSolvers/LineSearches.jl/blob/master/src/static.jl#L29
https://github.com/JuliaNLSolvers/LineSearches.jl/blob/master/src/initialguess.jl#L17

anriseth:
> Uhm... lol.. seems like it is already in InitialStatic, so basically this is done twice...

Hehe, very foresighted of me :grin:

PR #70, opened 2017-11-02
Create types for initial step length guess

### LINKED ISSUE #69: Create functionality for alphaguess
- Move alphaguess from Optim.
- Make it a type chosen by the user.
- Default all of them to alpha = 1 (Optim change. Except Accelerated and Momentum GD)

Initial ones to implement:
- Set alpha = 1
- Keep the previous alpha
- Extrapolation of the form from L-BFGS
- Converg `alphatry`

anriseth:
Fixes #69 

Currently missing: 
- [x] Documentation. I should list the different InitialGuess types in the README.
- [x] Extrapolate guess from L-BFGS implemented by @cortner. 

I'm not exactly sure how to handle the if-statement in L-BFGS `alphaguess`, where it checks whether `state.pseudo_iteration > 1`. Was the intention here to not use extrapolation when `L-BFGS` is reset due to non-positive-definite Hessian approximation, or is it fine as long as `state.f_x_previous` is finite? (@cortner if you have time, can you give quick feedback on this, please?)


anriseth:
I'll set up the Optim side whenever I've dealt with the extrapolation step. 
Then we can tag LineSearches, so that Optim can update its `alphaguess!` functionality.

src/hagerzhang.jl
```diff
@@ -486,3 +503,125 @@ function linefunc!(df,
...
+    ψ2::T         = 2.0
+    ψ3::T         = 0.1
+    αmax::T       = Inf
+    α0::T         = NaN # Initial alpha guess. NaN => algorithm calculates
+    verbose::Bool = false
+end
+
+
+function (is::InitialHagerZhang)(state, dphi0, df)
+    if isnan(state.f_x_previous) && isnan(is.α0)
```
pkofod: Is `isnan(state.f_x_previous)` used to check if this is the very first iteration? Alternatively, we could just pass the iteration number. It may be fine, I'm just curious if that is the purpose. If we keep it this way, you should probably add a comment here.

src/hagerzhang.jl
```diff
@@ -486,3 +503,125 @@ function linefunc!(df,
...
+end
+
+# Pick the initial step size (HZ #I1-I2)
+function _hzI12(alpha::T,
+                df,
+                x::Array,
+                s::Array,
+                xtmp::Array,
+                lsr::LineSearchResults,
+                psi1::Real = convert(T,0.2),
```
pkofod: Should these just be set in the function body? I mean, how are people going to set these, and more importantly: who are going to set these? :)

src/hagerzhang.jl
```diff
@@ -486,3 +503,125 @@ function linefunc!(df,
...
+        println("alpha guess (expand): ", alpha)
+    end
+    return alpha, mayterminate
+end
+
+# Generate initial guess for step size (HZ, stage I0)
+# TODO: deal with different types for the inputs
+function _hzI0(x::Array{T},
+               gr::Array,
+               f_x::Real,
```
pkofod: `f_x::T`?

src/initialguess.jl
```diff
@@ -0,0 +1,40 @@
...
+is scaled with the `l_2` norm of the step direction.
+"""
+@with_kw struct InitialStatic{T}
+    alpha::T = 1.0
+    scaled::Bool = false # Scales step. alpha ← min(alpha,||s||_2) / ||s||_2
+end
+
+function (is::InitialStatic)(state, dphi0, df)
+    state.alpha = is.alpha
+    if is.scaled == true && (ns = norm(state.s)) > zero(typeof(is.alpha))
```
pkofod: you have `typeof(is.alpha)` in `InitialStatic`, so you might as well use it
```
julia> struct A{T} end

julia> function (a::A{T})(x) where T
           println(convert(T, x))
       end

julia> a=A{Float64}()
A{Float64}()

julia> a(3)
3.0

julia> a(BigInt(3))
3.0
```
so you can simply add the type parameter 
```
funciton (is::InitialStatic{T})(state, dphi0, df) where T
...
```

src/initialguess.jl
```diff
@@ -0,0 +1,40 @@
...
+        # TODO: Type instability if there's a type mismatch between is.alpha and ns
+        state.alpha *= min(is.alpha, ns) / ns
```
pkofod: In most cases `eltype(state.s)` will be `typeof(is.alpha)` though, but yeah...

src/hagerzhang.jl
pkofod: (I know you didn't do this, I'm just asking since we're here now)

src/hagerzhang.jl
anriseth: Yes, this was my hack to check whether it's the first iteration.

src/hagerzhang.jl
```diff
@@ -115,7 +132,7 @@ function _hagerzhang!(df,
         phic, dphic = linefunc!(df, x, s, c, xtmp, true)
     end
     if !(isfinite(phic) && isfinite(dphic))
-        println("Warning: failed to achieve finite new evaluation point, using alpha=0")
+        Base.warn("Failed to achieve finite new evaluation point, using alpha=0")
```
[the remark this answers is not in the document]
anriseth: I somehow had it in my head that warn was not exported :o

src/hagerzhang.jl
anriseth: I was wondering whether that would be better as well

src/hagerzhang.jl
anriseth: Are you referring to `x`, `s`, `xtmp` and `lsr`, which are all in `state` from the `Optim` perspective, 
or to `psi1` etc. ?

I agree we can do some more invasive changes as we're already doing this :)

src/hagerzhang.jl
anriseth: I'll change it. 

Strictly speaking, however, `x` and `gr` are in different spaces, so there may be cases where we want different types / eltypes here.

pkofod: Yes, I know. Anyway, it shouldn't really matter.

src/hagerzhang.jl
pkofod: I'm talking about all the "optional" positional arguments. Not sure if the original author meant for these to be keywords or not, but since you can't set `psi1` anywhere, it might as well just be a variable set in the function body.

src/hagerzhang.jl
pkofod: Alright, it would be nice if there was a comment just mentioning this.

src/hagerzhang.jl
anriseth: The optional arguments can be set by the user when they call `InitialHagerZhang()`. The only option you can't set is the `iterfinitemax`. That was just me not thinking far enough, I'll add that as well.

anriseth: Actually, the value of `iterfinitemax` seems to be related to the type information, to keep the value of `x` finite. 

So I think there is no reason for the user to set it, and if there is it would be just as complicated for them to edit the source code as anything else.

anriseth:
I'm satisfied with this now. @pkofod let me know if you're OK with the changes (and the sparse documentation).

cortner:
sorry I am slow to respond. I am trying to remember what I did but can't find the original `alphaguess` implementation. Where should it be?

anriseth:
> I am trying to remember what I did but can't find the original alphaguess implementation. Where should it be?

I think you are referring to https://github.com/JuliaNLSolvers/Optim.jl/blob/master/src/utilities/perform_linesearch.jl#L30

It seems most likely that this extrapolation should work fine even if the approximate Hessian is no longer SPD (which is when `pseudo_iteration` is reset). So as long as `f_x_previous` has a correct value, I think this makes sense.

cortner:
Looking at that code, my only concern now is that this guess does not impose an upper bound on alpha. Other than that it looks fine and has worked very well for me in the past. 

ANyhow, your question was about `state.pseudo_iteration > 1`. I don't think I wrote that. If I did then I don't remember. After reset, I would normally use 1 for a well-scaled problem or otherwise one of the heuristics based on |grad|. There are probably some of those already implemented in Optim

anriseth:
> Looking at that code, my only concern now is that this guess does not impose an upper bound on alpha. 

Good point, I'll give the user an option to set a maximum (and minimum) alpha.

> After reset, I would normally use 1 for a well-scaled problem or otherwise one of the heuristics based on |grad|. There are probably some of those already implemented in Optim

I see. I think it should be possible for users to compose the different `Initial*` objects in a way that let's them do that, as long as the information needed is stored in the `state` object, `dphi0` or `df`.

src/initialguess.jl
```diff
@@ -70,7 +71,7 @@ function (is::InitialQuadratic{T})(state, dphi0, df) where T
         αguess = is.α0
     else
         αguess = 2.0 * (NLSolversBase.value(df) - state.f_x_previous) / dphi0
-        αguess = min(one(T), 1.01*αguess) # See Nocedal+Wright
+        αguess = min(is.αmax, 1.01*αguess) # See Nocedal + Wright, using is.αmax = 1.0
```
cortner: `min(amax, guess)` can only be a good choice when the problem is very well-scaled. What if the scaling is such that the optimal step-length is ~ 1e-2?

src/initialguess.jl
```diff
@@ -70,7 +71,7 @@ function (is::InitialQuadratic{T})(state, dphi0, df) where T
...
         αguess = max(is.αmin, αguess)
```
cortner: I don't think there should be a lower bound on the initial step-length. What if the problem is scaled such that 1e-2 is the optimal step-length? This would not be extraordinarily bad scaling at all. I generally prefer 
```
aguess = max(0.25*aprevious, guess)
```
if you want `amin`  in there, then the default should be much smaller I think.

cortner: For Newton methods I agree. For quasi-Newton I strongly disagree. The premise that 1.0 is a good step is generally only true in the asymptotic regime, which in many problems you never hit. But maybe in practise this is not as bad - or at least not very often. Maybe leave this for now and see if it causes any trouble.

src/initialguess.jl
```diff
@@ -0,0 +1,122 @@
...
+end
+
+function (is::InitialQuadratic{T})(state, dphi0, df) where T
+    if !isfinite(state.f_x_previous) || dphi0 ≈ zero(T)
+        # If we're at the first iteration
+        αguess = is.α0
+    else
+        αguess = 2.0 * (NLSolversBase.value(df) - state.f_x_previous) / dphi0
+        αguess = NaNMath.max(is.αmin, state.alpha*is.ρ, αguess)
+        αguess = NaNMath.min(is.αmax, 1.01*αguess) # See Nocedal + Wright, using is.αmax = 1.0
```
cortner: continuing from the previous conversation. Why not have a `snap2one` kind of thing here instead?

anriseth: You have almost convinced me that this could be a just-as-good approach. It achieves something similar to what N+W tries to do with `min(1.01aguess, amax)`, however, I'm not sure exactly what the difference will be in practice.

I also see from your defaults in the Optim code, with `snap2one=(0.75,Inf)`, which would be the same as setting `amax = 1.0`. Do you think those are good  `snap2one` default values, or shall I use something else?

Over the last week, I've gotten a much better appreciation for how important a good step length guess is. So I'm happy to discuss these properly before implementing some damaging default parameters. (Over the long term, optimizing over default parameters on a large test set would be very neat)

src/initialguess.jl
anriseth: > The premise that 1.0 is a good step is generally only true in the asymptotic regime, which in many problems you never hit.

My impression is that part of L-BFGS's big success is that it tends to work very well with a step length of 1.0, *so long as one implements the scaling of the inverse Hessian* (as I am attempting to do in https://github.com/JuliaNLSolvers/Optim.jl/pull/484 )

cortner: Good point. Once you scale the inverse hessian I am happy!

src/initialguess.jl
cortner: > I also see from your defaults in the Optim code, with snap2one=(0.75,Inf), which would be the same as setting amax = 1.0. Do you think those are good snap2one default values, or shall I use something else?

I'm not sure I set those default values, but either way this sounds ok.

> Over the last week, I've gotten a much better appreciation for how important a good step length guess is. So I'm happy to discuss these properly before implementing some damaging default parameters. 

Yes I found that as well when I implemented my first optimisation code. 

> (Over the long term, optimizing over default parameters on a large test set would be very neat)

I think that would be fantastic.

cortner:
From my perspective I'm happy with this PR, so don't wait for me to merge it in case I don't have time to review again.

anriseth:
This is ready now. @pkofod any final comments?

pkofod:
Perfect. Good to see you updated NEWS and README as well!

PR #71, opened 2017-11-05
Use NaNMath

anriseth:
In case something goes wrong during the interpolation step.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.6
 NLSolversBase 3.0
 Parameters
+NaNMath
--- a/src/LineSearches.jl
+++ b/src/LineSearches.jl
@@ -2,7 +2,7 @@ isdefined(Base, :__precompile__) && __precompile__()
 
 module LineSearches
 
-using Parameters
+using Parameters, NaNMath
 
 import NLSolversBase
 import Base.clear!
--- a/src/backtracking.jl
+++ b/src/backtracking.jl
@@ -73,7 +73,7 @@ function _backtracking!(df,
 
         # Shrink proposed step-size:
         if order == 2 || iteration == 1
-            # backtracking via interpolation:
+            # backtracking via quadratic interpolation:
             # This interpolates the available data
             #    f(0), f'(0), f(α)
             # with a quadractic which is then minimised; this comes with a
@@ -82,6 +82,7 @@ function _backtracking!(df,
             # of the function guarantees at least a backtracking factor rho.
             alphatmp = - (gxp * alpha^2) / ( 2.0 * (f_x_scratch - f_x - gxp*alpha) )
         else
+            # Backtracking via cubic interpolation
             alpha0 = lsr.alpha[end-1]
             alpha1 = lsr.alpha[end]
             phi0 = lsr.value[end-1]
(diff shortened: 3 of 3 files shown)


andreasnoack:
This causes the following problem
```julia
julia> NaNMath.min(ForwardDiff.Dual{Float64}(1), ForwardDiff.Dual{Float64}(1))
ERROR: StackOverflowError:
Stacktrace:
 [1] min(::ForwardDiff.Dual{Float64,Int64,0}, ::ForwardDiff.Dual{Float64,Int64,0}) at /Users/andreasnoack/.julia/packages/NaNMath/pEdac/src/NaNMath.jl:291 (repeats 80000 times)

julia> min(ForwardDiff.Dual{Float64}(1), ForwardDiff.Dual{Float64}(1))
Dual{Float64}(1)
```
which can happen with `autodiff=:forward`.

pkofod:
@anriseth do you remember what part "could go wrong"?

anriseth:
Hmm, no unfortunately not. I remember running into an issue where NaNs propagated through and the culprit may have been alphatmp in some weird edgecase.

We can either revert the Namath operations which are unlikely to affect as many people as the Forwarddiff problems, or add an explicit check on NaN instead 
(I believe the problem was alphatmp, and not alpha, rhohi, or rholo)

andreasnoack:
I tried to change `NaNMath.min` and `max` in `Backtracking` locally and I immediately got gradients full of `NaN`s so I probably hit the issue that made you do this PR.

anriseth:
Ah, that's annoying. Do any of the other line search methods work for you?
I assume `Static` would work. We could make a vanilla backtracking as well, to get something, potentially, a little better than `Static` but which support Duals.

pkofod:
The other searches don’t really work for Andreas, because they don’t have the initial finiteness backtracking step (it should just be included in the others as well)

anriseth:
> because they don’t have the initial finiteness backtracking step

I may have misunderstood, but I thought they do?

https://github.com/JuliaNLSolvers/LineSearches.jl/blob/master/src/hagerzhang.jl#L142
https://github.com/JuliaNLSolvers/LineSearches.jl/blob/master/src/morethuente.jl#L216
https://github.com/JuliaNLSolvers/LineSearches.jl/blob/master/src/static.jl#L20

pkofod:
Oh that was added... did you try morethuente Andreas?

andreasnoack:
I've locally defined `DiffRules` for `NaNMath.min/max` but then I hit the `NaN` issue right after. The culprit seems to be
https://github.com/JuliaNLSolvers/LineSearches.jl/blob/fae0768f4a79d1fec637a53104c24d4761312abe/src/backtracking.jl#L32.
Since `maxstep` is initialized to `Inf` we get things like
```julia
julia> Inf/ForwardDiff.Dual{Float64}(1.0, 0.0)
Dual{Float64}(Inf,NaN)
```
Initially, I thought I could work around this by flipping the `NANSAFE_MODE_ENABLED` flag in `ForwardDiff` but it didn't work, see https://github.com/JuliaDiff/ForwardDiff.jl/pull/179#issuecomment-462521308.

I guess it might be difficult to get correct behavior for `Dual`s when we hit `Inf`s. Maybe another solution could be to pick set `ls.maxstep` to a large but finite value.

andreasnoack:
I can of course just set `maxstep` but it might be difficult to figure out.

anriseth:
If I remember correctly `maxstep` is something @cortner found useful (correct me if I'm wrong). By setting the default to `Inf` we thought it wouldn't make any difference to people anyway, but you've proved that wrong :p 

Two possible ways around this:
1. If you've defined DiffRules for `NanMath.min`, then would it work to make the line
`α_0 = NaNMath.min(α_0, NaNMath.min(alphamax, ls.maxstep / norm(s, Inf)))` ?
2. We can explicitly check for `NaN` and skip the `min` operation when there are `NaN`s (or does that cause issues elsewhere?)

If you remove the line checking for `maxstep` locally, does the optimization run to convergence?

cortner:
Correct - I introduced this As a stabilising mechanism.

Could one just check whether maxstep is Inf and in that case skip this operation?

andreasnoack:
Regarding the initial comment then I think it should be fixed by https://github.com/JuliaDiff/DiffRules.jl/pull/31.

Regarding the other issue then https://github.com/JuliaDiff/ForwardDiff.jl/pull/386 might help here but it's still an option that is off by default.

I'm not sure what the better solution is but at least I've been able to get things working now for my use case by specifying a finite `maxstep`.

PR #73, opened 2017-11-17
Halve step in MoreThuente if initial step generates nonfinite values

anriseth:
Also adds some `@inbounds` steps to BackTracking


pkofod:
Is this from from the original ?

anriseth:
I haven't looked at the original paper or code, however, I hit this problem
when running "Exponential". It follows the same approach as hagerzhang and
backtracking to ensure the "initial" step length provides finite function
values.

PR #75, opened 2017-11-22
Fixes to MoreThuente

anriseth:
After looking through the MoreThuente code, I found that the original translation from the MATLAB code has a bug in it. There are several lines in `cstep` where a value `s = max(theta, dgx,dg)` is calculated. In the original [FORTRAN](http://www.cs.umd.edu/~oleary/LBFGS/FORTRAN/linesearch.f) and [MATLAB](https://github.com/andrewssobral/mtt/blob/master/libs/poblano_toolbox_1.1/cstep.m#L170)  codes, however, this is set to `s = max(abs(theta), abs(dgx), abs(dg))`.

I caught this when an optimization problem I was running broke down. It is a high-dimensional problem, and I can't replicate it in a test as the input values to the line search are shown in pretty-mode, and small changes in the eight decimal place mean that the problem does not appear...

There also seems to be a weird choice in the original codes where they calculate
`gamma = s* sqrt( (theta/s)^2 - (dg/s) * (dgx / s))`. As `s >= 0`, this should be equivalent (modulo floating points) to `gamma = sqrt( theta^2 - dg * dgx)`. I have made the change, as it omits the potentially problematic division by `s`.

In addition, I have moved the finite-value check from #73 outside the of loop, so that it does not interfere with the algorithm in itself, but only ensures that the initial step does not mess things up.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,6 @@
+# LineSearches v3.2.2 release notes
+- Ensure finite initial values and fix typos in original MoreThuente translation from MATLAB/FORTRAN implementations. See [#75](https://github.com/anriseth/LineSearches.jl/pull/75)
+
 # LineSearches v3.2.0 release notes
 - Introduce initial step length procedures. See [#70](https://github.com/anriseth/LineSearches.jl/pull/70)
 
--- a/src/morethuente.jl
+++ b/src/morethuente.jl
@@ -496,7 +496,7 @@ function cstep(stx::Real, fx::Real, dgx::Real,
       bound = true
       theta = 3 * (fx - f) / (stp - stx) + dgx + dg
       # Original code from FORTRAN implementation
-      # s = maxabs(theta, dgx, dg)
+      # s = maximum(abs,[theta, dgx, dg])
       # gamma = s * sqrt((theta / s)^2 - (dgx / s) * (dg / s))
       # The s cancels if you move it inside the sqrt?????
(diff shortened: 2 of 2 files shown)


pkofod:
Good catch... and did it work after you fixed this?

anriseth:
> Good catch... and did it work after you fixed this?

Yes, it now converges, whilst previously it generated NaN step guesses.

I have also compared behaviour with a without the s term in the gamma. This does change the number of finding evaluations in two of the optimisations I ran (these a corner cases with step directions almost orthogonal to the descent direction). The calculated difference for gamma are order 1e-15 or smaller (sometimes zero), so these algorithms are not always stable to floating point operation errors...

pkofod:
Alright, looks good to me then. They probably did it for a reason, if it's a floating point reason, maybe @simonbyrne has stumbled across something like it before?

anriseth:
I'll get in contact with the people who were involved in the implementations of the original codes, and see if they can remember (originally from 1983 and 1991, so we'll see...)

anriseth:
Dianne O'Leary pointed out to me that `s` is a scaling to prevent under- and overflow in the calculation of `theta^2` and `dgx * dg`. It's probably best to keep it that way.

I'll merge and tag this on Tuesday, if nobody has any objections.

pkofod:
Unless you have a specific reason not to, just squash+merge and tag now (or whenever you feel like it).

PR #77, opened 2018-01-17
Use vecnorm

anriseth:
Currently, some of the algorithms call `norm` on the step directions or gradients, however, if we work with input objects such as `Array{Float64,3}`, then there is no norm function. To handle this, I changed to calling `vecnorm`. 

@antoine-levitt @cortner Is `vecnorm` a sensible norm to use when you optimize with inputs being complex numbers and/or from manifolds? Or do we need to support a user-defined norm function for these cases (in the future)?

BackTracking and HagerZhang already use `vecnorm`.

### DIFF
--- a/src/initialguess.jl
+++ b/src/initialguess.jl
@@ -14,7 +14,7 @@ end
 
 function (is::InitialStatic{T})(state, dphi0, df) where T
     state.alpha = is.alpha
-    if is.scaled == true && (ns = norm(state.s)) > zero(T)
+    if is.scaled == true && (ns = vecnorm(state.s)) > zero(T)
         # TODO: Type instability if there's a type mismatch between is.alpha and ns
         state.alpha *= min(is.alpha, ns) / ns
     end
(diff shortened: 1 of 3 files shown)


antoine-levitt:
`vecnorm` is definitely better than `norm` for this purpose. For complex numbers you want `vecnorm` (because Optim is working internally with a 2 x whatever array) anyway. For manifolds this is only called (hopefully) on tangent vectors and so it's the same as non-manifolds. In all of the cases I'm familiar with the Frobenius norm (`vecnorm`) is more natural than the matrix norm (`norm`). I wish Base called `vecnorm` `norm` and `norm` `matrix_norm` or something like that: `norm` is O(n^3) (I think), so it's very easy to put an `@assert norm(diff) < 1e-6` when writing code and wondering why your code suddenly takes ages to run.

User-defined norms (really, inner products) should instead go as preconditioning.

cortner:
I agree with everything @antoine-levitt said.

anriseth:
Great, thank you both for your feedback. Then this change shouldn't mess up anybody's complex number or Manifold optimizations (but rather fix any issues).

PR #78, opened 2018-01-21
Move to AbstractArray.

pkofod:


anriseth:
Did you got an issue with this anywhere? Or was it to ensure we won't in the future?

pkofod:
It's to allow array of arrays as inputs https://discourse.julialang.org/t/optim-arguments-higher-dimensional-arrays-and-gradient-specification/8500/7

PR #80, opened 2018-02-26
Kill LineSearchResults.

pkofod:
I also want to unify all the `phiprime` `dphi` etc names. Is `phi_0, dphi_0` fine, or is `phi0, dphi0` better?


src/backtracking.jl
```diff
@@ -53,38 +55,40 @@ function _backtracking!(df,
     n = length(x)
 
     # read f_x and slope from LineSearchResults
-    @inbounds f_x = lsr.value[end]
-    @inbounds gxp = lsr.slope[end]
+    @inbounds f_x = phi0
+    @inbounds gxp = dphi0
```
anriseth: Remove `@inbounds`

src/backtracking.jl
```diff
@@ -96,40 +100,37 @@ function _backtracking!(df,
...
-            alphatmp = - (gxp * alpha^2) / ( 2.0 * (f_x_scratch - f_x - gxp*alpha) )
+            alphatmp = - (gxp * alpha1^2) / ( 2.0 * (f_x_scratch - f_x - gxp*alpha1) )
         else
             # Backtracking via cubic interpolation
-            @inbounds alpha0 = lsr.alpha[end-1]
-            @inbounds alpha1 = lsr.alpha[end]
-            @inbounds phi0 = lsr.value[end-1]
-            @inbounds phi1 = lsr.value[end]
+            @inbounds phi0 = values[end-1]
+            @inbounds phi1 = values[end]
```
anriseth: This seems to be the only use of `values`, so maybe we don't need to create an array here?

src/backtracking.jl
```diff
@@ -96,40 +100,37 @@ function _backtracking!(df,
...
                 alphatmp = gxp / (2.0*b)
             else
-                discr = max(b^2-3*a*gxp, zero(alpha))
+                discr = max(b^2-3*a*gxp, zero(Tα))
                 alphatmp = (-b + sqrt(discr)) / (3.0*a)
             end
         end
-        alphatmp =  NaNMath.min(alphatmp, alpha*rhohi) # avoid too small reductions
-        alphatmp = NaNMath.max(alphatmp, alpha*rholo) # avoid too big reductions
+        alphatmp =  NaNMath.min(alphatmp, alpha1*rhohi) # avoid too small reductions
```
anriseth: Remove the unnecessary space (I know it was there from before, woops)

src/deprecate.jl
```diff
@@ -39,3 +39,31 @@ function alphatry(alpha::T,
 
     LineSearches._hzI12(alpha, df, x, s, xtmp, lsr, psi1, psi2, psi3, alphamax, verbose)
 end
+
+
+(ls::Static)(df, x, s, x_scratch, lsr, alpha, mayterminate) = (ls::Static)(df, x, s, x_scratch, alpha)
```
anriseth: Should we add a deprecation message here?

src/morethuente.jl
```diff
@@ -137,22 +137,18 @@
     f_tol::T = 1e-4 # c_1 Wolfe sufficient decrease condition
     gtol::T = 0.9   # c_2 Wolfe curvature condition (Recommend 0.1 for GradientDescent)
     x_tol::T = 1e-8
-    stpmin::T = 1e-16
-    stpmax::T = 65536.0
+    alphamin::T = 1e-16
```
anriseth: This seems to have been chosen to be `eps(Float64)`. I guess we can generalise?

src/morethuente.jl
```diff
@@ -172,16 +171,16 @@ function _morethuente!(df,
...
         throw(ArgumentError("Invalid parameters to morethuente"))
     end
 
 
     # read finit and slope from LineSearchResults
-    f = lsr.value[end]
-    dginit = lsr.slope[end] # dot(gradient(df),s)
-    if dginit >= 0.0
+    f = phi0
+    dphi0 = dphi0 # dot(gradient(df),s)
```
anriseth: This line can be removed

src/morethuente.jl
```diff
@@ -205,43 +204,43 @@ function _morethuente!(df,
...
+    dgy = dphi0
 
     # START: Ensure that the initial step provides finite function values
     # This is not part of the original FORTRAN code
-    if !isfinite(stp)
-        stp = one(T)
+    if !isfinite(alpha)
+        alpha = one(T)
     end
     stmin = stx
```
anriseth: Maybe we should update `stx`, `sty`, `stmin` and `stmax` to also be called something with `alpha`?

Note the awkward situation where `stpmin` and `stmin` are different variables, so we can't rename both to `alphamin` ... 

`stmin` and `stmax` are used as inputs to `cstep`, so maybe we can call them `alphacmin`, `alphacstepmin` or something like that?

src/morethuente.jl
```diff
@@ -205,43 +204,43 @@ function _morethuente!(df,
...
-        @. x_new = x + stp*s
+        alpha = 0.5*alpha
+        @. x_new = x + alpha*s
         f = NLSolversBase.value_gradient!(df, x_new)
         gdf = NLSolversBase.gradient(df)
         nfev += 1 # This includes calls to f() and g!()
 
-        # Make stpmax = (3/2)*stp < 2stp in the first iteration below
-        stx = (7/8)*stp
+        # Make alphamax = (3/2)*alpha < 2alpha in the first iteration below
```
anriseth: This was a typo by me. I meant to say "Make `stmax`", which is different from `stpmax` ...

src/morethuente.jl
```diff
@@ -468,7 +467,7 @@ end # function
 function cstep(stx::Real, fx::Real, dgx::Real,
```
anriseth: Rename `stx`, `sty`, `stp`, `stpc`, `stpf` etc. as well?

src/strongwolfe.jl
```diff
@@ -27,7 +27,8 @@ function _strongwolfe!(df,
                       x::AbstractArray{T},
                       p::AbstractArray{T},
                       x_new::AbstractArray{T},
-                      lsr::LineSearchResults{T},
+                      phi_0,
```
anriseth: Should we add types here?

src/types.jl
```diff
@@ -30,3 +30,8 @@ mutable struct LineSearchException{T<:Real} <: Exception
     alpha::T
     lsr::LineSearchResults
 end
+
+mutable struct NewLineSearchException{T<:Real} <: Exception
```
anriseth: What are you planning to use this for? :)

src/deprecate.jl
pkofod: I should have noted it. I will change all these definitions to proper deprecations,yeah.

src/types.jl
pkofod: That was not supposed to be committed like this :) I was just experimenting with exceptions :)

pkofod:
Thanks! I'll have a look tomorrow, and hopefully finish HagerZhang

pkofod:
Could you have a look again @anriseth ?

src/hagerzhang.jl
```diff
@@ -276,119 +281,131 @@ function _hagerzhang!(df,
                 println("Linesearch: secant failed, using bisection")
             end
             c = (A + B) / convert(T, 2)
-            # phic = phi(gphi, c) # TODO: Replace
-            phic, dphic = linefunc!(df, x, s, c, xtmp, true)
-            @assert isfinite(phic) && isfinite(dphic)
-            push!(lsr, c, phic, dphic)
-            # ia, ib = update(phi, lsr, iA, iB, length(lsr), philim) # TODO: Pass options
-            ia, ib = update!(df, x, s, xtmp, lsr, iA, iB, length(lsr), philim, display)
+            # phi_c = phi(gphi, c) # TODO: Replace
```
anriseth: Is this line necessary anymore? Seems like it has been replaced already ...

src/morethuente.jl
```diff
@@ -468,7 +467,7 @@ end # function
...
                sty::Real, fy::Real, dgy::Real,
                stp::Real, f::Real, dg::Real,
-               bracketed::Bool, stpmin::Real, stpmax::Real)
+               bracketed::Bool, alphamin::Real, alphamax::Real)
```
anriseth: As you're staying with `stp` in this function, I think it's better to keep using `stpmin` instead of `alphamin`.

src/deprecate.jl
anriseth: Deprecation message now, or are you planning to do that at a later stage / just leave it until we implement the one-dimensional LineSearches?

src/morethuente.jl
pkofod: I'm fine with alpha

src/deprecate.jl
pkofod: I'll do it before this is merged

anriseth:
Great, merge it whenever you like :)

PR #83, opened 2018-03-03
Fix initial guesses to work without lsr and a method for static to match

pkofod:


pkofod:
@anriseth would you be okay with v4.0.0 simply breaking the old API? The problem is that HagerZhang has all sort of special needs, so it's sort of a pain to unify APIs while dropping LSR support. We can also go the fully backwards compatible way, but that requires temporarily having an extra field in the method states types in Optim.

Since we should be the only users of this I'd be fine with this hard break, but it's up to you.

anriseth:
Let's be breaking. As has been mentioned multiple places, until we change the interface here to optimise one-dimensional functions there isn't a big use case for others to use the package.

anriseth:
It has dawned on me that LineSearchResults originally may have been written with HagerZhang in mind. It seems to be using it much more heavily than any of the other algorithms. 

Similarly `mayterminate` seems to be a `HagerZhang` thing - so we'll have to think about how to deal with that going forward as well.

anriseth:
Oh, and the changes look good to me :+1:

pkofod:
Yes that’s what I realized as well (wrt HZ).

PR #84, opened 2018-03-04
Update backtracking

pkofod:
@anriseth 

this removes the unnecessary vector and renames some stuff.

### DIFF
--- a/src/backtracking.jl
+++ b/src/backtracking.jl
@@ -9,85 +9,67 @@ there exists a factor ρ = ρ(c₁) such that α' ≦ ρ α.
 This is a modification of the algorithm described in Nocedal Wright (2nd ed), Sec. 3.5.
 """
 @with_kw struct BackTracking{TF, TI}
-    c1::TF = 1e-4
-    rhohi::TF = 0.5
-    rholo::TF = 0.1
+    c_1::TF = 1e-4
+    ρ_hi::TF = 0.5
+    ρ_lo::TF = 0.1
     iterations::TI = 1_000
     order::TI = 3
     maxstep::TF = Inf
 end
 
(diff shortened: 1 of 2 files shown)


anriseth:
Looks like we're getting closer to a univariate interface :)

In terms of consistency across the algorithms, will \alpha_2 be easy to use as the representative of the "most recent guess" of the step length everywhere? (It's not really that crucial, but if we could use the same variable name for all algorithms that would be neat)

pkofod:
> In terms of consistency across the algorithms, will \alpha_2 be easy to use as the representative of the "most recent guess" of the step length everywhere? (It's not really that crucial, but if we could use the same variable name for all algorithms that would be neat)

Would you rather \alpha_a, \alpha_b, \alpha_c, \alpha_d, .... for first, second, third, fourth, ... ?

anriseth:
I like numbers better. Maybe we can just put a comment in the codes, for the user's benefit of understanding which variable is intended to be returned? (I know they can just look for return statements, but this helps with chronology when reading the code)

pkofod:
So do you prefer `_1`, `_2`, ... ? In strongwolfe they're labelled `\alpha_i`, `alpha_iminus1`.

anriseth:
I think I'll retract my wish for "consistency" :p 

Looking at HagerZhang, it seems difficult to unify the variable names in the different algorithms, so let's just go with whatever makes the most sense for each algorithm individually. As you are going through them now, I say you go with what is more readable to you.

There is also an advantage to using the same variables as used by the authors of the papers describing the algorithms, as it is easier to compare the papers/books with the code we have.

pkofod:
> There is also an advantage to using the same variables as used by the authors of the papers describing the algorithms, as it is easier to compare the papers/books with the code we have.

I think this may be the best argument. Consistency where it makes sense, but fall back to paper / text book notation.

PR #85, opened 2018-03-05
Univariate-ize backatracking and strongwolfe.

pkofod:
This shows how it could look.

There is a slightly annoying thing here, and that is that there will be a minor increase in cost if there's a an fg! method available, so we might want to allow for passing in phi, dphi, phi_dphi?

### DIFF
--- a/src/hagerzhang.jl
+++ b/src/hagerzhang.jl
@@ -525,6 +525,8 @@ otherwise, we select according to procedure I1-2, with starting value α0.
 end
 
 function (is::InitialHagerZhang)(state, phi_0, dphi_0, df)
+
+
     if isnan(state.f_x_previous) && isnan(is.α0)
         # If we're at the first iteration (f_x_previous is NaN)
         # and the user has not provided an initial step size (is.α0 is NaN),
@@ -557,6 +559,9 @@ function _hzI12(alpha::T,
                 alphamax::Real = convert(T, Inf),
                 verbose::Bool = false) where T
 
+
+     ϕ = make_ϕ(df, x_new, x, s)
+
     # Prevent values of `x_new` that are likely to make
     # ϕ(x_new) infinite
     iterfinitemax::Int = ceil(Int, -log2(eps(T)))
@@ -564,14 +569,14 @@ function _hzI12(alpha::T,
     alphatest = psi1 * alpha
     alphatest = min(alphatest, alphamax)
 
-    @. x_new = x + alphatest * s
-    phitest = NLSolversBase.value!(df, x_new)
+    phitest = ϕ(alphatest)
 
     iterfinite = 1
     while !isfinite(phitest)
         alphatest = psi3 * alphatest
-        @. x_new = x + alphatest * s
-        phitest = NLSolversBase.value!(df, x_new)
+
+        phitest = ϕ(alphatest)
+
         iterfinite += 1
         if iterfinite >= iterfinitemax
             return zero(T), true
--- a/src/hagerzhang.jl
+++ b/src/hagerzhang.jl
@@ -118,7 +118,7 @@ function _hagerzhang!(df,
     ϕ, dϕ, ϕdϕ = make_ϕ_dϕ_ϕdϕ(df, x_new, x, s)
 
     # Prevent values of `x_new` that are likely to make
-    # df.f(x_new) infinite
+    # ϕ(x_new) infinite
     iterfinitemax::Int = ceil(Int, -log2(eps(T)))
     alphas = [T(0.0)] # for bisection
     values = [phi_0]
@@ -558,7 +558,7 @@ function _hzI12(alpha::T,
                 verbose::Bool = false) where T
 
     # Prevent values of `x_new` that are likely to make
-    # df.f(x_new) infinite
+    # ϕ(x_new) infinite
     iterfinitemax::Int = ceil(Int, -log2(eps(T)))
 
     alphatest = psi1 * alpha
--- a/src/hagerzhang.jl
+++ b/src/hagerzhang.jl
@@ -100,7 +100,7 @@ end
 function _hagerzhang!(df,
                      x::AbstractArray{T},
                      s::AbstractArray{T},
-                     xtmp::AbstractArray{T},
+                     x_new::AbstractArray{T},
                      phi_0,
                      dphi_0,
                      c::Real,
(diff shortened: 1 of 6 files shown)


pkofod:
I should note that current I haven't had to update https://github.com/JuliaNLSolvers/Optim.jl/pull/543 so no functional changes

anriseth:
> There is a slightly annoying thing here, and that is that there will be a minor increase in cost if there's a an fg! method available, so we might want to allow for passing in phi, dphi, phi_dphi?

Yes, I'd say we include phi_dphi in the algorithms that require it. If we also have some sort of `UnivariateDifferentiable` function then we can do `ls(df) = ls(df.phi, df.dphi, df.phi_dphi)` as well, which may make the interface simpler for Optim?

src/static.jl
```diff
@@ -1,48 +1,37 @@
...
+# for generic use. We use ls.alpha instead.
 
 """
 `Static`: defines a static linesearch, i.e. with fixed step-size. E.g., initialise
 with `Static(alpha = 0.3141)` for fixed step-size 0.3141. Default is 1.0.
 
 You can also make this independent of the size of the step `s`, by using
 `Static(scaled = true)`.
-This will then use a steps-size alpha ← min(ls.alpha,||s||_2) / ||s||_2
+This will then use a steps-size alpha ← min(alpha,||s||_2) / ||s||_2
```
anriseth: "step-size" (my fault)

pkofod:
It's taking shape. Of course the final step is to remove `df` completely from the signature, but this shows how it would work. Still works on the optim PR.

src/hagerzhang.jl
```diff
@@ -512,28 +505,6 @@ function bisect!(df,
     return ia, ib
 end
 
-# Define one-parameter function for line searches
-function linefunc!(df,
```
anriseth: Ah, nice :)

src/hagerzhang.jl
```diff
@@ -115,8 +115,10 @@ function _hagerzhang!(df,
                      psi3::Real = convert(T,0.1),
                      display::Integer = 0) where T
 
-    # Prevent values of `xtmp` that are likely to make
-    # df.f(xtmp) infinite
+    ϕ, dϕ, ϕdϕ = make_ϕ_dϕ_ϕdϕ(df, x_new, x, s)
+
+    # Prevent values of `x_new` that are likely to make
+    # df.f(x_new) infinite
```
anriseth: "make \phi(x_new) infinite"

src/hagerzhang.jl
```diff
@@ -586,21 +557,21 @@ function _hzI12(alpha::T,
                 alphamax::Real = convert(T, Inf),
                 verbose::Bool = false) where T
 
-    # Prevent values of `xtmp` that are likely to make
-    # df.f(xtmp) infinite
+    # Prevent values of `x_new` that are likely to make
+    # df.f(x_new) infinite
```
anriseth: "\phi(x_new) infinite"

src/hagerzhang.jl
anriseth: (Nice that we don't need this anymore)

pkofod: This seems to have been the precursor for our approach here.

pkofod:
The only annoying thinig is that HagerZhang is strictly worse now, because it allocates every round. I need to go through it again to see if we cannot just have 4 or 5 \phi's, vals, and slopes

pkofod:
I need to make HZ non-allocating, but can you review this for merge-worthiness @anriseth ?

pkofod:
I will make the final transition in a separate PR

src/hagerzhang.jl
```diff
@@ -586,21 +557,21 @@ function _hzI12(alpha::T,
...
+    # ϕ(x_new) infinite
     iterfinitemax::Int = ceil(Int, -log2(eps(T)))
 
     alphatest = psi1 * alpha
     alphatest = min(alphatest, alphamax)
 
-    @. xtmp = x + alphatest * s
-    phitest = NLSolversBase.value!(df, xtmp)
+    @. x_new = x + alphatest * s
+    phitest = NLSolversBase.value!(df, x_new)
```
anriseth: Can you use \phi(alphatest) here?

src/hagerzhang.jl
```diff
@@ -586,21 +557,21 @@ function _hzI12(alpha::T,
...
 
     iterfinite = 1
     while !isfinite(phitest)
         alphatest = psi3 * alphatest
-        @. xtmp = x + alphatest * s
-        phitest = NLSolversBase.value!(df, xtmp)
+        @. x_new = x + alphatest * s
+        phitest = NLSolversBase.value!(df, x_new)
```
anriseth: Ditto \phi(alphatest)

pkofod:
What do we do about _hzI0 ? That explicitly requires the gradient of the objective and the actual point.

anriseth:
> What do we do about _hzI0 ? That explicitly requires the gradient of the objective.

`_hzI0` is only used to select an initial guess, so it doesn't impact the line search algorithm per se. Maybe move it (and InitialHagerZhang) to initialguess.jl and deal with it separately?

Hmm, I guess we need to decide whether we wish to treat the initial step length methods from the viewpoint of the optimization algorithm using `df`, with the flexibility that brings, or whether it should also adhere to the univariate-framework.

pkofod:
> Hmm, I guess we need to decide whether we wish to treat the initial step length methods from the viewpoint of the optimization algorithm using df, with the flexibility that brings, or whether it should also adhere to the univariate-framework.

So you'd be fine with all the initial guessing being "separate" from the actual searches, such that the initial guess CAN be used, but might not be used.

anriseth:
> So you'd be fine with all the initial guessing being "separate" from the actual searches, such that the initial guess CAN be used, but might not be used.

Yes. My philosophy is that the line search algorithms can be given any initial step length. The `Initial*` codes are strategies that encode information from a "separate" application, in determining what is the best guess for the initial step length. These strategies should allow more flexible interfaces, independent of how the line search algorithms go about determining the final step length.

pkofod:
> Yes. My philosophy is that the line search algorithms can be given any initial step length. The Initial* codes are strategies that encode information from a "separate" application, in determining what is the best guess for the initial step length. These strategies should allow more flexible interfaces, independent of how the line search algorithms go about determining the final step length.

Good, then we can leave these things as they are for now.

PR #86, opened 2018-03-16
Univariatize even more.

pkofod:
@anriseth that `mayterminate` is so annoying... I have proposed a change but I don't really like it, as it requires intial guesses and the actual line searches to communicate through `state.mayterminate` that should then hold the same pointer as in HagerZhang... would it make sense to pass in the linesearch as an argument to the initial guesses ? I think that's the only robust solution.


anriseth:
> @anriseth that mayterminate is so annoying... 

I think it may be valuable if we look through the HZ paper (or the code we have) and see what the point of `mayterminate` is, and why the decision on its value is made in the initial guess.

HZ sort-of breaks the "univariatization" approach to line searches that people have requested (and I agree will be useful for generalization). It does seem like HZ is "too specialized" towards optimization, but I'll have to look through the paper before I stand behind my claims here.

anriseth:
I propose the following, let me know if that's satisfactory.
- We remove `mayterminate` from `state`, and add it to `HagerZhang`. 
- We can consider passing the LineSearch to the InitialGuess methods, and then tell `_hzI12` to update `mayterminate` only if the line search is of type `HagerZhang`.


I've had a look at the code John wrote, as well as the Hager FORTRAN code. This is definitely something which is a `HagerZhang` only functionality, and the only `InitialGuess` that can produce a `mayterminate` true is the QuadStep initial step length guess from Hager and Zhang's `CG_DESCENT` paper. 

Also, `mayterminate` seems to be the same as `QuadOK` in the Hager code. It is used more places in that code than by John, so I think our line search is not exactly replicating Hager and Zhang's (but it still seems to do quite well, so let's not try to mess about with it :p )

pkofod:
@anriseth is `dphi_0_previous` only ever used in `InitialConstantChange`? Then we might want to find a better way of communicating that it's the first iteration.

anriseth:
> @anriseth is dphi_0_previous only ever used in InitialConstantChange? Then we might want to find a better way of communicating that it's the first iteration.

Yes it is only used there, and it would definitely be great if we can somehow store dphi0_previous in the InitialConstantChange rather than in state. (That way we could also use NaN to represent the first iteration)

anriseth:
Getting closer and closer :)

What remains for this to be fully transitioned now? Is it only dealing with dphi_previous ?

pkofod:
That and giving all the line searches a generic user facing api (the problem I mentioned on gitter)

Issue #88, opened 2018-03-24
Missing tolerance in `isapprox`


mohamed82008:
https://github.com/JuliaNLSolvers/LineSearches.jl/blob/27dd9b674eb9742490529aba5e9988a01f109ace/src/morethuente.jl#L281

The above line should have an absolute tolerance in light of https://discourse.julialang.org/t/approximate-equality/8952/4.

anriseth:
Yeah, that makes sense. Ideally we'd set this tolerance as a parameter in
the MoreThuente type/function call, so that users can adapt it to their
optimization tolerance (for example). If you wish to make a PR that
hardcodes the tolerance, feel free to do so :)

I'd put the default to either eps(T) or, if possible, something emulating
the default tolerances we currently have in Optim.

PR #89, opened 2018-03-24
Arbitrary precision support

mohamed82008:
Also fixed https://github.com/JuliaNLSolvers/LineSearches.jl/issues/88.


src/backtracking.jl
```diff
@@ -76,19 +76,19 @@ function (ls::BackTracking)(df, x::AbstractArray{T}, s::AbstractArray{T},
...
 
             if isapprox(a, zero(a))
-                α_tmp = dϕ_0 / (2.0*b)
+                α_tmp = dϕ_0 / (T(2)*b)
             else
                 # discriminant
                 d = max(b^2 - 3*a*dϕ_0, zero(Tα))
                 # quadratic equation root
-                α_tmp = (-b + sqrt(d)) / (3.0*a)
+                α_tmp = (-b + sqrt(d)) / (3*a)
```
anriseth: T(3) ?

src/deprecate.jl
```diff
@@ -1 +1 @@
-# Deprecation warnings
+# Deprecation warnings
```
anriseth: Github complains about a missing newline here ( I think)

src/hagerzhang.jl
```diff
@@ -200,8 +200,7 @@ function _hagerzhang!(df,
             if c > alphamax
                 c = (alphamax + cold)/2
                 if display & BRACKET > 0
-                    println("bracket: exceeding alphamax, bisecting: alphamax = ", alphamax,
-                            ", cold = ", cold, ", new c = ", c)
+                    println("bracket: exceeding alphamax, bisecting: alphamax = ", alphamax, ", cold = ", cold, ", new c = ", c)
```
anriseth: This line has become a bit long now (just leave it as it was if there are no changes?)

src/hagerzhang.jl
```diff
@@ -312,7 +311,7 @@ function satisfies_wolfe(c::T,
                          sigma::Real) where T<:Number
     wolfe1 = delta * dphi_0 >= (phi_c - phi_0) / c &&
                dphi_c >= sigma * dphi_0
-    wolfe2 = (2.0 * delta - 1.0) * dphi_0 >= dphi_c >= sigma * dphi_0 &&
+    wolfe2 = (2 * delta - 1) * dphi_0 >= dphi_c >= sigma * dphi_0 &&
```
anriseth: Do you have a consistent approach that I have missed for when you use `T(1)` `T(2)` and when you just use `2`?

src/morethuente.jl
```diff
@@ -167,12 +167,12 @@ function (ls::MoreThuente)(df,
     # Check the input parameters for errors.
     #
 
-    if  alpha <= 0.0 || f_tol < 0.0 || gtol < 0.0 ||
-        x_tol < 0.0 || alphamin < 0.0 || alphamax < alphamin || maxfev <= 0
+    if  alpha <= 0 || f_tol < 0 || gtol < 0 ||
```
anriseth: Again, why are you using 0 some places and `zero(T)` other places?

src/morethuente.jl
```diff
@@ -278,7 +278,7 @@ function (ls::MoreThuente)(df,
         f, dg = ϕdϕ(alpha)
         nfev += 1 # This includes calls to f() and g!()
 
-        if isapprox(dg, 0.0)
+        if isapprox(dg, 0, atol=eps(T))
```
anriseth: Can you add a comment here to say that we may want to add the `atol` value to MoreThuente in future, please?

src/morethuente.jl
```diff
@@ -380,8 +380,8 @@ function (ls::MoreThuente)(df,
         #
 
         if bracketed
-            if abs(sty - stx) >= 0.66 * width1
-                alpha = stx + 0.5 * (sty - stx)
+            if abs(sty - stx) >= T(33)/50 * width1
```
anriseth: Is this better than doing `T(2)/3` ?

src/morethuente.jl
```diff
@@ -452,15 +452,17 @@ function cstep(stx::Real, fx::Real, dgx::Real,
...
 
    #
    # Check the input parameters for error
    #
 
    if (bracketed && (alpha <= min(stx, sty) || alpha >= max(stx, sty))) ||
-        dgx * (alpha - stx) >= 0.0 || alphamax < alphamin
-      throw(ArgumentError("Minimizer not bracketed"))
+        dgx * (alpha - stx) >= 0 || alphamax < alphamin
+        throw(ArgumentError("Minimizer not bracketed"))
```
anriseth: I think you want to remove two spaces here

src/morethuente.jl
```diff
@@ -505,7 +507,7 @@ function cstep(stx::Real, fx::Real, dgx::Real,
    # the cubic step is taken, else the quadratic step is taken
    #
 
-   elseif sgnd < 0.0
+   elseif sgnd < 0
       info = 2
       bound = false
       theta = 3 * (fx - f) / (alpha - stx) + dgx + dg
```
anriseth: 0 vs zero(T) and T(3) vs 3?

src/morethuente.jl
```diff
@@ -640,9 +642,9 @@ function cstep(stx::Real, fx::Real, dgx::Real,
    alpha = alphaf
    if bracketed && bound
       if sty > stx
-         alpha = min(stx + 0.66 * (sty - stx), alpha)
+         alpha = min(stx + T(33)/50 * (sty - stx), alpha)
```
anriseth: Advantage vs `T(2)/3` ?

src/strongwolfe.jl
```diff
@@ -98,8 +98,8 @@ function zoom(a_lo::T,
               x::AbstractArray,
               p::AbstractArray,
               x_new::AbstractArray,
-              c_1::Real = 1e-4,
-              c_2::Real = 0.9) where T
+              c_1::Real = T(1)/10^4,
```
anriseth: Is there a difference to the compiler whether we do `T(1)`*10^(-4)` and and `T(1)/10^4` ?

src/strongwolfe.jl
```diff
@@ -143,7 +143,7 @@ function zoom(a_lo::T,
                 return a_j
             end
 
-            if ϕprime_a_j * (a_hi - a_lo) >= 0.0
+            if ϕprime_a_j * (a_hi - a_lo) >= 0
```
anriseth: 0 vs zero(T)?

src/strongwolfe.jl
```diff
@@ -161,9 +161,9 @@ function interpolate(a_i1::Real, a_i::Real,
                      ϕ_a_i1::Real, ϕ_a_i::Real,
                      dϕ_a_i1::Real, dϕ_a_i::Real)
     d1 = dϕ_a_i1 + dϕ_a_i -
-        3.0 * (ϕ_a_i1 - ϕ_a_i) / (a_i1 - a_i)
+        3 * (ϕ_a_i1 - ϕ_a_i) / (a_i1 - a_i)
```
anriseth: 3 vs T(3)?

src/backtracking.jl
mohamed82008: No need. Integer and float operations always use the float type as output.
```julia
julia> 3*Float32(1)
3.0f0

julia> 3*BigFloat(1)
3.000000000000000000000000000000000000000000000000000000000000000000000000000000
```

src/hagerzhang.jl
mohamed82008: Oh sure didn't even realize I edited it!

src/hagerzhang.jl
mohamed82008: There is a bit of inconsistency I noticed, partly because I thought `T(1)` was necessary before. The main criteria though is to never allow division of integers since that goes by default to `Float64`. I will double check it tomorrow.

src/morethuente.jl
mohamed82008: I will make it more consistent.

src/morethuente.jl
mohamed82008: Nothing, I just didn't want to change the code semantics!

src/strongwolfe.jl
mohamed82008: `10^(-4)` is an error. `(10.)^-4` or `1e-4` are `Float64`s, and I am trying to avoid automatic conversion to any undesired precision. I think with inter-procedural constant propagation, all these constants will compile away in v0.7 so performance will not be harmed, supposedly.

src/morethuente.jl
mohamed82008: Sure.

src/strongwolfe.jl
mohamed82008: Same effect if the RHS is of type `T`.

src/strongwolfe.jl
mohamed82008: My tests show that `zero(T)` is faster with `BigFloat`, I will change it.

src/morethuente.jl
mohamed82008: I will make it consistent using `T(3)`.

src/hagerzhang.jl
mohamed82008: The criteria is if the integer gets multiplied by a float, then I don't need to wrap it since the result is going to be of the correct type whether I wrap it or not, so I opt for the cleaner look of simply using the integer.

src/morethuente.jl
anriseth: That makes sense. Can you add a small comment about it originally being 0.66 and that the new value is for supporting other types?

If you have time, you can check whether the More Thuente paper says anything about the value, and whether it "should" be 2/3.

mohamed82008:
I tried following the following rules for dealing with integer literals:

1. If an integer is divided by another integer, wrap the first one with `T` to stop conversion to `Float64`. Same thing should be done for `log`, `exp` or any unary function that returns `Float64` by default.
2. If an integer is used in a logical operation `<`, `>`, etc. make sure the terms compared are all of the same type, since this is faster according to my experiments.
3. If an integer is multiplied, added or subtracted to another float, leave it as is to avoid cluttering since automatic conversion happens anyways. This is slower than converting when benchmarking using `HigherPrecision` and `BigFloat` but changing it is cumbersome so I will do it next week.

src/morethuente.jl
mohamed82008: Hmm I actually changed it to 2/3 in the latest commit but I will check the paper.

src/backtracking.jl
```diff
@@ -51,7 +51,7 @@ function (ls::BackTracking)(df, x::AbstractArray{T}, s::AbstractArray{T},
     while !isfinite(ϕx_1) && iterfinite < iterfinitemax
         iterfinite += 1
         α_1 = α_2
-        α_2 = 0.5*α_1
+        α_2 = (T(1)/2)*α_1
```
pkofod: Might aswell write `T(0.5)` as 0.5 is exactly representable as a double float, so there should be no loss of accuracy, or maybe

```
α_2 = α_1/2
```
is even better?

src/backtracking.jl
pkofod: Yes it should be fine

src/morethuente.jl
```diff
@@ -222,13 +222,13 @@ function (ls::MoreThuente)(df,
     iterfinite = 0
     while (!isfinite(f) || !isfinite(dg)) && iterfinite < iterfinitemax
         iterfinite += 1
-        alpha = 0.5*alpha
+        alpha = (T(1)/2)*alpha
```
pkofod: I guess you could also just divide by 2 here?

src/morethuente.jl
```diff
@@ -380,8 +380,8 @@ function (ls::MoreThuente)(df,
         #
 
         if bracketed
-            if abs(sty - stx) >= 0.66 * width1
-                alpha = stx + 0.5 * (sty - stx)
+            if abs(sty - stx) >= (T(2)/3) * width1
```
pkofod: Are the parentheses necessary here? Why not just `T(2)/3 * width1`?

src/backtracking.jl
mohamed82008: Yup could do that. I was in semi-robotic mode just replacing all decimals by their equivalent!

src/morethuente.jl
mohamed82008: Agreed.

src/morethuente.jl
mohamed82008: I was being elaborate to avoid any re-ordering. If the compiler re-orders it to `T(2 ) * width1 / 3` it may stop constant propagation and inlining the constant, so just an extra precaution, but it seems unnecessary.

pkofod: Okay, I don't know much about the behavior of the new constant proagation, and I think it might actually be a bit magic, so maybe it's better to be safe than sorry!

pkofod:
Can you add some tests with some of the types that now work? You simply add the packages in the `/test/REQUIRE` to avoid the whole project depending on it.

mohamed82008:
You mean make separate testing package and just use it in runtests.jl?

pkofod:
No. If you want to test `ArbFloats.jl` you can simply make an `ArbFloats` entry in the `test/REQUIRE` file and then in `test/runtests.jl` you put `using ArbFloats` at the top. Then `ArbFloats` will be installed when `Pkg.test("Optim")` is run, and deleted after it is finished.

pkofod:
> Oh I see, I will try to do it as soon as I can.

Perfect! I promise to be responsive in return :)

pkofod:
Thanks a  lot @mohamed82008 . This was a battle, I look forward to finishing the war! I merged to avoid conflicts because I'm actively working in this repo, but please, I urge you to come back and add some of the tests I asked for earlier.

mohamed82008:
Awesome, thanks! I will do the tests hopefully this weekend at least partially.

pkofod:
> Awesome, thanks! I will do the tests hopefully this weekend at least partially.

Anything is welcome, thanks!

PR #90, opened 2018-04-05
Further improvements to inner methods

pkofod:


src/backtracking.jl
```diff
@@ -88,15 +87,15 @@ function (ls::BackTracking)(ϕ, x::AbstractArray{T}, s::AbstractArray{T}, α_0::
...
+            div = one(T) / (α_1^2 * α_2^2 * (α_2 - α_1))
             a = (α_1^2*(ϕx_1 - ϕ_0 - dϕ_0*α_2) - α_2^2*(ϕx_0 - ϕ_0 - dϕ_0*α_1))*div
             b = (-α_1^3*(ϕx_1 - ϕ_0 - dϕ_0*α_2) + α_2^3*(ϕx_0 - ϕ_0 - dϕ_0*α_1))*div
 
             if isapprox(a, zero(a), atol=eps(T))
                 α_tmp = dϕ_0 / (2*b)
             else
                 # discriminant
-                d = max(b^2 - 3*a*dϕ_0, zero(Tα))
+                d = max(b^2 - 3*a*dϕ_0, zero(T))
```
anriseth: Do you want to use `T(0)` here as well?

src/backtracking.jl
```diff
@@ -106,7 +105,7 @@ function (ls::BackTracking)(ϕ, x::AbstractArray{T}, s::AbstractArray{T}, α_0::
 
         # enforce a maximum step alpha * s (application specific, default is Inf)
         α_1 = α_2
-        α_2 = min(α_tmp, maxstep / vecnorm(s, Inf))
+        α_2 = min(α_tmp, maxstep) # / vecnorm(s, Inf))
```
anriseth: Do you plan to somehow include the `vecnorm` value in other ways?

pkofod: Hm, good question... Why would it be scaled by s in the first place? `s` is constant throughout the optimization, so in my mind it makes more sense to pass a `maxstep`, and if the user then wants to calculate that from the sup-norm then they can do that outside of `ls(...)`?

anriseth: @cortner , you added the division by `vecnorm(s, Inf)` in https://github.com/JuliaNLSolvers/LineSearches.jl/pull/43

Can you advise us on what the intention was here, and whether it is useful to preserve the behaviour in the way we call the new univariatized LineSearches from Optim?

cortner: The intention is to restrict the actual step length not the step length parameter alpha

pkofod: > Note also the comment that default maxstep = Inf in which case this will not be used.

This is not the point, though, the point is that since we're now passing `phi` functions to the line searches, it is no longer needed to pass things like `x`, `x_tmp` and  `s` to the line searches. `s` is certainly *not*  changing throughout a line search, so I think it's much better to handle the `maxstep` calculation outside of the actual line search, which also means that it is easier to reuse the functionality across line searches. I have no intentions of changing the behavior you'd experience, but I want to do it in another way in the code.

cortner: I see - then there should be a maxstep parameter in the optimisation method which translates to a maxalpha parameter, or in backtracking just the initial alpha

cortner: So this line should be removed altogether

pkofod: Something like it has to remain , I’ll return to it soon :)

PR #95, opened 2018-04-08
Arbitrary precision tests and HagerZhang{T} and BackTracking{TF} constructors

mohamed82008:
I added some basic tests for `Float32`, `Float64`, `DoubleFloat64` from HigherPrecision.jl, and `BigFloat`. The tests are based off the tests/initial.jl file and only check that outputs are of the correct type and not NaN, so they are pretty basic. I think the real tests will be easier and better to do at the Optim side when it is also made to support arbitrary precision. If you think there is more that I should do, please let me know.

While I was at it, I realized that HagerZhang and BackTracking have a second parameter, and it is usually irrelevant to the user, so I made 2 new constructors of the syntax `BackTracking{TF}(args...; kwargs...) where TF = BackTracking{TF, Int}(args...; kwargs...)` and its like for `HagerZhang`. This means the syntax for making a `HagerZhang` instance of precision `T` will be simply `HagerZhang{T}(...)`.


mohamed82008:
I am not sure why it says 13 commits but meh!

pkofod:
Hm, tests seem to fail on all platforms

test/arbitrary_precision.jl
```diff
@@ -0,0 +1,142 @@
...
+        @testset "Arbitrary precision - initial step guess: $($T)" begin
+            f(x) = vecdot(x, x)
+            function g!(out, x)
+                out[:] = 2x
+            end
+
+            x = $T[-1, -1]
+            df = NLSolversBase.OnceDifferentiable(f,g!,x)
+
+            phi_0 = NLSolversBase.value_gradient!(df, x)
```
anriseth: value_gradient! now returns the gradient as well, so you'll have to update this to something like 
```jl
phi_0, grtmp = NLSolversBase.value_gradient!(df, x)
```

test/arbitrary_precision.jl
```diff
@@ -0,0 +1,142 @@
...
+            @test !isnan(phi_0)
+            @test phi_0 isa $T
+            grtmp = NLSolversBase.gradient(df)
```
anriseth: This line is then no longer needed

mohamed82008:
I will look into this.

anriseth:
Are you planning to add more code to this PR @mohamed82008 , or is the only thing missing to update the two lines I commented on? (If so, I can just do it for you and merge this )

mohamed82008:
I apologize for the delay, this has not been up my priority list for a while. I am not planning to add more code, so if you want to take care of it, please go ahead. Thank you!

anriseth:
Win32 overflows because of HigherPrecision. Is that expected?

anriseth:
Thanks:)

PR #97, opened 2018-04-23
Remove scaling from Static 

anriseth:
This should remove the final barrier for us to fully univariatise the line searches.

The `scaled=true` feature can still be achieved by using `InitialStatic(scaled=true)` together with `Static()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ in `Optim`.
 * `MoreThuente` (From the algorithm in More and Thuente, 1994)
 * `BackTracking` (Described in Nocedal and Wright, 2006)
 * `StrongWolfe` (Nocedal and Wright)
-* `Static` (Simply takes a given step length, default 1.0)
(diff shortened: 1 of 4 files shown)


anriseth:
@pkofod have you got a good suggestion for deprecating this? I removed the `alpha` field from Static as well, so if anybody uses that things will crash.

anriseth:
@pkofod have you got a good suggestion for deprecating this? I removed the `alpha` field from Static as well, so if anybody uses that things will crash.

anriseth:
There was a bug in the original implementation of InitialStatic... That teaches me to write tests in a hurry.

anriseth:
I tried to deprecate the new static, but it didn't work.

anriseth:
Okay, now Static is deprecated in a hacky way. We should remove this hack by v6.0, ref #99.

PR #101, opened 2018-05-14
Ensure function value is finite for Static step

anriseth:
We should probably add some sort of debug / verbosity flag here to warn if this happens. (Future work :p )


pkofod:
> (Future work :p )

Hey, that means open an issue! ;)

anriseth:
I hope this doesn't mess up #90 too much.

PR #106, opened 2018-05-30
Remove x as an input in the inner methods.

pkofod:


pkofod:
@anriseth good?

anriseth:
Looking at this it has made me wonder why we think `s` is fine to keep, but not `x`. 

It makes more sense to me to have both or neither. If one wants to know what `s` is, then understanding what its relative size is compared to `x` may also be of value. I prefer neither. 

(Sorry)

pkofod:
Point taken. I think I'd rather avoid introducing them again later. Let's just keep them in the interface.

anriseth:
> I think I'd rather avoid introducing them again later.

I think this is unlikely but fair enough.

We could also write shorter versions of the functions that just pass `Nothing` (or whatever it's called in 0.7) to the linesearches:
```jl
ls(\phi, \alpha) = ls(\phi, \alpha, Nothing, Nothing)
```

pkofod:
Alright, I'm removing both.

anriseth:
> Alright, I'm removing both. 

You know who to blame if it turns out we need them :blush: :worried:

pkofod:
> You know who to blame if it turns out we need them blush worried

Oh, I will... I will....

We just need to solve the problem of alphamax in backtracking then

anriseth:
Can we introduce some sort of `PrepareLineSearch` method?

I imagine this doing the following:
- Decide initial step length
- Determine / decide maximum step length (and possible minimum step length)
- Set other parameters (for example, the IPNewton line search requires a value that depends on the Hessian of the Lagrangian, so we could provide this as some parameter input )

pkofod:
G.. d... windows...

PR #107, opened 2018-05-31
Fix exports

anriseth:
We don't have LineSearchResults anymore, but we do have LineSearchException.


anriseth:
Maybe something happened when they released Julia 0.6.3?

PR #111, opened 2018-06-01
Add Julia 0.7 to CI

anriseth:
This also changes `ls` to `linesearch` in the custom optimizer example.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ os:
   - osx
 julia:
   - 0.6
+  - 0.7
   - nightly
 matrix:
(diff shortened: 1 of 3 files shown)


pkofod:
Do you want it to succeed on v0.7, or did you just want to include it to test against alpha?

anriseth:
> Do you want it to succeed on v0.7, or did you just want to include it to test against alpha?

I thought it would succeed already.

Hmm, it seems Travis didn't recognise 0.7

And funnily enough, win32 0.7 passes but win64 0.7 does not; the tables have turned.

PR #112, opened 2018-06-01
Remove dphi_0_previous from State and move it into InitialConstantCha…

pkofod:
…nge.

This removes that last annoying mix of Optim and LineSearches (I think?)


anriseth:
> This removes that last annoying mix of Optim and LineSearches (I think?)

:tada:

PR #113, opened 2018-06-01
Introduce docs

anriseth:
Here's an attempt at introducing documentation in combination with Literate examples.

I propose that we introduce docs in three steps (if you are okay with making docs this way @pkofod )
1. Merge this and check out what the docs look like
2. Move the Optim examples in the README to a Literate.jl example
3. Possibly write a larger manual for each of the line search methods (and step length guesses)

### DIFF
--- a/docs/generate.jl
+++ b/docs/generate.jl
@@ -1,0 +1,13 @@
+# generate examples
+import Literate
+
+EXAMPLEDIR = joinpath(@__DIR__, "src", "examples")
+GENERATEDDIR = joinpath(@__DIR__, "src", "examples", "generated")
+for example in filter!(r"\.jl$", readdir(EXAMPLEDIR))
+    input = abspath(joinpath(EXAMPLEDIR, example))
+    script = Literate.script(input, GENERATEDDIR)
+    code = strip(read(script, String))
+    mdpost(str) = replace(str, "@__CODE__" => code)
+    Literate.markdown(input, GENERATEDDIR, postprocess = mdpost)
+    Literate.notebook(input, GENERATEDDIR, execute = true)
+end
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,0 +1,33 @@
+using Documenter, LineSearches
+
+# Generate examples
+include("generate.jl")
+
+GENERATEDEXAMPLES = [joinpath("examples", "generated", f) for f in (
+    "customoptimizer.md",)]
+
+# Build documentation.
+makedocs(
+    format = :html,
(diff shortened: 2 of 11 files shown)


anriseth:
I put the Documenter script call in `runtests.jl`.

Let me know if you prefer that it is added as a separate call in `.travis.jl` in the same fashion as we do in `Optim.`

pkofod:
Won’t it deploy in all builds then?

anriseth:
> Won’t it deploy in all builds then?

Hmm, I don't know. I just saw that this is how they do it in `JuAFEM`. 

@KristofferC @fredrikekre is there a benefit in calling `Documenter` from `runtests.jl` rather than in `.travis.jl`?

fredrikekre:
> is there a benefit in calling Documenter from runtests.jl rather than in .travis.jl?

The benefit is that if the doc-build fail (for any reason, for example doctests) then the CI run is marked as a fail. Steps that fail in `after_success` on travis does not mark the build as a fail.) The other advantage is that you can put Documenter in `test/REQUIRE` and it will be installed automatically withouth manual `Pkg.add`'s.

>Won’t it deploy in all builds then?

We do this:
https://github.com/KristofferC/JuAFEM.jl/blob/fb3cc17a57f54ac89a137d80fa85328e082dda0b/test/runtests.jl#L21
to only run docs on one worker.

anriseth:
Thanks for the input :)

I use the same Travis build check as JuAFEM, so it sounds like the setup in this PR works fine.

fredrikekre:
Need to add a SSH-key to Travis (https://juliadocs.github.io/Documenter.jl/latest/man/hosting/)
```
Warning: DOCUMENTER_KEY environment variable missing, unable to deploy.
  Note that in Documenter v0.9.0 old deprecated authentication methods were removed.
  DOCUMENTER_KEY is now the only option. See the documentation for more information.
```

pkofod:
should be added now

fredrikekre:
Cool: https://julianlsolvers.github.io/LineSearches.jl/latest/

I just wanted to see how it worked out with Literate! 😄

anriseth:
Yes, it seems to work well :D 

Haha, I kept the line from JuAFEM, woops.

> and then run
```jl
using JuAFEM
```

pkofod:
Seems to work just fine! Awesome. I'll have @iewaij add his coming Optim examples using `Literate.jl` as well

pkofod:
> Haha, I kept the line from JuAFEM, woops.

Let's just add JuAFEM to require then ;)

PR #114, opened 2018-06-02
Create more docs examples and link from README

anriseth:
This moves the README examples of how to use LineSearches with Optim to the docs.

There's an awkward issue with running CI tests and docs deployment with Literate examples that use packages that depend on LineSearches:
- If we introduce a breaking change in LineSearches, it must be tagged before Optim
- Running Optim + LineSearches in CI / Travis deployment will use incompatible versions of Optim and LineSearches. 

I have therefore added a filter that only generates and tests the Optim-dependent examples when we are running locally.


anriseth:
Maybe we should run Documenter makedocs each time we call Pkg.test, so that that we remember to update the output of the "static" examples when the output changes.

Then we could move `deploydocs` to `runtests.jl` but only call it when we're on a particular Travis build.

anriseth:
Hmm, so that fails because the code blocks in the "static" markdown pages I have added run in a Documenter `@example` block, which runs the code and provides the output.

anriseth:
I think this gives a good balance between local testing of examples and preventing Optim-LineSearches breakage to mess up CI and documentation deployment.

PR #118, opened 2018-06-05
Fix \phi_0 and d\phi_0 values

anriseth:
The values of ϕ_0 and dϕ_0 are ϕ(0), dϕ(0), not ϕ(\alpha_0)).

It is a bit confusing to use α_0  as the initial guess, because it means that we think ϕ_0 refers to the value at the initial guess. Maybe we should change all α0 / α_0's to αinitial?

### DIFF
--- a/src/backtracking.jl
+++ b/src/backtracking.jl
@@ -23,10 +23,10 @@ function (ls::BackTracking)(df::AbstractObjective, x::AbstractArray{T}, s::Abstr
     ϕ, dϕ = make_ϕ_dϕ(df, x_new, x, s)
 
     if ϕ_0 == nothing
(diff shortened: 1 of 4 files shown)


pkofod:
Hm, I don't think I agree with this change. phi_0 zero is meant to be the initial phi (and dphi as well), so those calculations were just meant to calculate the values if they weren't provided already. Or am I missing something?

anriseth:
No, phi_0 and dphi_0 that the line stretch algorithms use are the values of phi at alpha=0. That is what Optim passes to the line searches.

pkofod:
Ah, right, sorry, it's the numbers used to check armijo/wolfe/.../conditions ... good catch :)

PR #125, opened 2018-06-20
Test Optim examples

anriseth:
Next time we hit a breaking change in LineSearches we'll have to remove these tests (and possibly Optim from REQUIRE) again. A bit of a faff, but more tests > fewer tests.

### DIFF
--- a/docs/generate.jl
+++ b/docs/generate.jl
@@ -3,7 +3,8 @@ import Literate
 
 # TODO: Remove items from `SKIPFILE` as soon as they run on the latest
 # stable `Optim` (or other dependency)
-ONLYSTATIC = ["optim_linesearch.jl", "optim_initialstep.jl"]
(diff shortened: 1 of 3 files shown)


anriseth:
Good or bad idea, @pkofod ?

pkofod:
Possibly a good idea as long as it doesn't mess something up I can't imagine right now.

anriseth:
I'll merge tomorrow if we can't think of anything (fingers crossed for Pkg3 saving the day if we've forgotten something). 

On the bright side, it increases the test coverage by 5 percentage points :)

pkofod:
As long as it doesn’t mess up JuliaCIbot and we have to tag ten different versions to publish a new version, I’m fine with this :)

anriseth:
The future will tell :grin:

PR #126, opened 2018-07-13
Updates for Julia 0.7-beta

dawbarton:
I've had a go at updating LineSearches for Julia 0.7-beta. Mostly it's replacing `vecdot` with `dot` and `vecnorm` with `norm` since they are now equivalent as far as I can tell from the deprecation messages and docstrings. There are a few other changes but nothing major.

As far as I can tell the tests pass except for the ones that depend on external packages that haven't been updated (e.g., Optim and HigherPrecision).

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.6
+julia 0.7-beta
 NLSolversBase 5.0
 Parameters
 NaNMath
--- a/docs/src/examples/customoptimizer.jl
+++ b/docs/src/examples/customoptimizer.jl
@@ -28,11 +28,11 @@ function gdoptimize(f, g!, fg!, x0::AbstractArray{T}, linesearch,
     ϕ(α) = f(x .+ α.*s)
     function dϕ(α)
         g!(gvec, x .+ α.*s)
-        return vecdot(gvec, s)
+        return dot(gvec, s)
     end
     function ϕdϕ(α)
         phi = fg!(gvec, x .+ α.*s)
-        dphi = vecdot(gvec, s)
+        dphi = dot(gvec, s)
         return (phi, dphi)
     end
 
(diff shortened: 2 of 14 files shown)


REQUIRE
```diff
@@ -1,4 +1,4 @@
-julia 0.6
+julia 0.7-beta
```
anriseth: Can you use -alpha here?

REQUIRE
```diff
@@ -1,4 +1,4 @@
...
 NLSolversBase 5.0
```
anriseth: This should probably be updated when we release the 0.7 version of NLSolversBase (https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/75 )

pkofod:
> What changes have been made to Julia that changes the floating point calculations involved?

I don't have too much time to check right now, but it could be that something like `norm(x, 2)` now uses a different summation order (maybe it loops, maybe it uses `sum`) because of the norm/vecnorm changes. This is just speculation, but similar changes to other base functions could cause something like this as well.

REQUIRE
anriseth: Or are we supposed to use `beta2`? I don't know what is recommended.

dawbarton: Personally I'd say use the latest version of 0.7 since there should be no reasons for people not to upgrade. (I can understand people needing to stay on 0.6 for legacy reasons but 0.7.0-alpha?!)

dawbarton:
I've found the problem with the test and, fortunately, it's not a problem with LineSearches.

Try the line 
```@test [9.9999995462402447e-01, 9.9999990907139824e-01] ≈ [1.0,1.0] atol=1e-7```
(just load the Test package, nothing else) in Julia 0.6.3 (**passes**) and in Julia 0.7.0-beta2 (**fails**). The numbers just come from the output of the last iteration in the optimiser (which are identical in both versions of Julia). So it's purely a change in how the test is being performed.

anriseth:
> So it's purely a change in how the test is being performed.

Okay, that's good. Thanks for looking into it :)

I've updated your PR in #127 with you as a commiter, so I'll merge that one instead.

PR #127, opened 2018-07-15
Require Julia 0.7

anriseth:
I just want to see what femtocleaner has to say here (Ref  #126 )

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
   - linux
   - osx
 julia:
-  - 0.6
+  - 0.7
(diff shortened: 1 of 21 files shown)


anriseth:
There were lots of deprecation warnings from HigherPrecision because `T(0)` no longer defaults to `convert(T, 0)`. 
I have therefore changed all lines calling `T(<number>)` to `convert(T, <number>)`. It does not look as nice, but the alternative would have to be to rely on all number types defining `T(x::Number) = convert(T, x)` ...

PR #129, opened 2018-09-12
Update for 1.0, follow CI scripts from NLSolversBase

anriseth:


pkofod:
Seems to be a problem with higherprecision?

anriseth:
Yes, it has been deprecated in favour of DoubleFloats.jl, but that opened another can of worms. https://github.com/JuliaMath/DoubleFloats.jl/issues/18 (default values set to `Inf` became `NaN` with `DoubleFloats`)

Issue #135, opened 2018-10-31
Bracket initialization - bug in HagerZhang? 


mohamed82008:
https://github.com/JuliaNLSolvers/LineSearches.jl/blob/1f40bf319887b77f66f89154763b5d49924894a7/src/hagerzhang.jl#L194

If I understand correctly, `ia` here should be 1 as per the paper which initializes the bracket to be [0, c_j].

Issue #134, opened 2018-10-31
Concerns about Hager Zhang initialization


mohamed82008:
I have a few questions/concerns about the initialization of HagerZhang:
1. If we start at or close to the minimizer, `dphi0` will be close to 0, so it is possible this will error https://github.com/JuliaNLSolvers/LineSearches.jl/blob/1f40bf319887b77f66f89154763b5d49924894a7/src/initialguess.jl#L252.
2. In the paper, the quadratic stepping is optional and when it fails, the algorithm falls back on using an initial guess that is a multiple of the previous step size. Isn't this a more reasonable thing to do rather than error above?
3. If I understand correctly, this line https://github.com/JuliaNLSolvers/LineSearches.jl/blob/1f40bf319887b77f66f89154763b5d49924894a7/src/initialguess.jl#L266 is not in the original paper.
4. While I understand the rationale behind the while loop in https://github.com/JuliaNLSolvers/LineSearches.jl/blob/1f40bf319887b77f66f89154763b5d49924894a7/src/initialguess.jl#L228 which was used to ensure that `phitest` is finite, possibly a concern with log-barrier methods, but this guarantee is gone in https://github.com/JuliaNLSolvers/LineSearches.jl/blob/1f40bf319887b77f66f89154763b5d49924894a7/src/initialguess.jl#L268. Maybe do a similar loop here?

mohamed82008:
Suggestions:
1. Remove the error and fallback on the next branch.
2. Check that `phitest` is finite at the end.
3. Make QuadStep an optional parameter.

Beside the fixes proposed in https://github.com/JuliaNLSolvers/LineSearches.jl/issues/133. If you approve, a PR will be next.

mohamed82008:
Another place where starting close to convergence may cause a problem https://github.com/JuliaNLSolvers/LineSearches.jl/blob/1f40bf319887b77f66f89154763b5d49924894a7/src/hagerzhang.jl#L119.

pkofod:
I have been aware about the staring close to a solution for a while, but never fixed it. A pr is very much welcome

pkofod:
This is brilliant, thanks. I think they will give fewer issues that relate to weird line search errors on well-behaved problems in the future!

PR #136, opened 2018-11-01
[WIP] Fix HagerZhang bugs

### LINKED ISSUE #133: Initial HagerZhang: norm(gr) -> norm(gr)^2
https://github.com/JuliaNLSolvers/LineSearches.jl/blob/1f40bf319887b77f66f89154763b5d49924894a7/src/initialguess.jl#L292

In the paper, the divisor is `norm(gr)^2`

mohamed82008:
This PR closes #133, #134, and #135. There is one error I am getting in LineSearches tests which seems unrelated to the changes made. Optim tests pass. I will look into this further. Also, I need to check if other places error when the search direction is not a descent direction and try to make it robust to computational error when starting at the optimal solution. Finally, someone please comment on my use of `mayterminate` since I don't understand how it will be used apart from checking when the quadratic step is taken.

### DIFF
--- a/src/hagerzhang.jl
+++ b/src/hagerzhang.jl
@@ -116,8 +116,10 @@ function (ls::HagerZhang)(ϕ, ϕdϕ,
     if !(isfinite(phi_0) && isfinite(dphi_0))
         throw(ArgumentError("Value and slope at step length = 0 must be finite."))
     end
-    if dphi_0 >= zeroT
+    if dphi_0 >= eps(T) * abs(phi_0)
         throw(ArgumentError("Search direction is not a direction of descent."))
+    elseif dphi_0 >= 0
+        return zeroT, phi_0
     end
 
     # Prevent values of x_new = x+αs that are likely to make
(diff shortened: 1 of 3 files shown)


mohamed82008:
I think this is good to go with the exception of some tests failing. The tests maybe depending on the old behavior? Does that make them good tests? Any help there is appreciated. Optim tests pass, so that's a sign that I didn't totally screw up.

pkofod:
I think that test is probably just check that behavior hasn't changed, but it seems quite likely that your changes come with changes to that exact step length. I Don't have time right now, but I'll review asap.

anriseth:
> The tests are maybe depending on the old behavior? Does that make them good tests?

Most are just to flag if we unintentionally change behaviour and are not good tests. They were added long after the implementation of the algorithms so little thought has been put into them. Feel free to update them if you believe the new behaviour is the correct one ;)

mohamed82008:
All good now.

pkofod:
Thanks! I believe code coverage is really bugged in julia these days..

pkofod:
I've tagged a bugfix release https://github.com/notifications

mohamed82008:
Awesome thanks for the quick merge!

PR #143, opened 2020-07-22
Fix exceptions in hagerzhang.jl

gwater:
LineSearchException expects two arguments, throw() expects only one


pkofod:
Hadn't seen this one. Just fixed it in another commit, but thanks!

PR #160, opened 2022-05-09
Updated NaNMath compatibility to v1.x

truedichotomy:


briochemc:
FWIW I think the issue here is that Appveyor.jl installs Julia v1.5 over there: 

https://github.com/JuliaCI/Appveyor.jl/blob/c4e6e502719dca34bd92c0127206ffe64fc0c0c1/bin/install.ps1#L31. 

You might be able to fix this by making a PR to Appveyor.jl. 

Alternatively, you could apply some band-aid here and enforce the use of Julia v1.6 here: 

https://github.com/JuliaNLSolvers/LineSearches.jl/blob/deb5db05d5aeaa97a218daba963ac3f50422c8c0/appveyor.yml#L3-L4

But personally I think it might be best to set this repo up with GitHub actions instead of Appveyor, following, e.g., https://github.com/julia-actions/julia-runtest.

pkofod:
Sorry for dragging my feet. I'm fixing this.

Issue #158, opened 2022-03-24
Bump NaNMath


GregPlowman:
Is it possible to bump `NaNMath` to version 1 ?

sethaxen:
Also it seems CompatHelper has been deactivated, or it would have opened a PR to do so.

rasmushenningsson:
bump :slightly_smiling_face:

pkofod:
Sorry sorry sorry... I'll do it

rasmushenningsson:
Thank you!

PR #161, opened 2022-05-09
Changed Julia to v1.6 and NaNMath to v1

truedichotomy:
Updating just NaNMath to v1 causes build failure for Julia v1. Since 1.6 is now the LTS version, make sense to up the version requirement for LineSearches as well.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "LineSearches"
 uuid = "d3d80556-e9d4-5f37-9878-2ab0fcc64255"
-version = "7.1.1"
+version = "7.1.2"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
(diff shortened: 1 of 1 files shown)


ederag:
The `appveyor` CI run failed because it had only `Julia Version 1.5.4`, which forced NaNMath to 0.3.
Could it be run again ?
Pinning NaNMath down looks like the root cause of this [Pluto downgrade](https://discourse.julialang.org/t/need-help-to-fix-a-pluto-downgrade-when-adding-another-package/82698).

HenriDeh:
Up. NaNMath is used by a lot of other packages and is often restricted to 0.3 if any optimization package is also in the deps. So it would be nice to update this requirement to support 1.0.

ederag:
Indeed. This is tracked in #158.

pkofod:
Sorry, I'm getting it done now.

PR #174, opened 2024-01-28
HZ: avoid convergence if bisected

### LINKED ISSUE #173: Suboptimal choice of alpha in Hager-Zhang
I've encountered a weird issue with alpha value selection in Hager-Zhang. Let me illustrate. Here are the values computed during bracketing:
```julia
alphas = [0.0, 1.0, 0.5, 0.25, 0.11797881082031754, 0.05898940541015877, 0.029494702705079385, 0.014747351352539692, 0.007373675676269846]
values = [32867.0, 33089.907756650115, 33027.619212962294, 32749.607153359673, 33107.9648525175, 33056.34280854859, 32990.02494986772, 32936.18207166336, 32831.66557482673]
slopes = [-171068.0, -29730.999177119797, -74242.25565689849, 191429.29678267744, -7900.963612183026, -51707.79133751766, -103642.35349017732, -139136.2036028941, 182648.13853509305]
```
The search ends on this line https://github.com/JuliaNLSolvers/LineSearches.jl/blob/ded667a80f47886c77d67e8890f6adb127679ab4/src/hagerzhang.jl#L287 , with `iA=1`, `ia=1`, `ib=4` and `iB=9`. Clearly, out of these four, `iA` is the worst choice (highest cost value) but nevertheless it's the value returned by the procedure. My function isn't particularly well-behaved but shouldn't the line search try a bit harder to pick the best value? I mean, it could just look at `ib` and `iB` instead of just returning `iA` in that line. What's worse here, the returned alpha is 0 which throws off the direction selection algorithm I'm using.

timholy:
`secant2` calls `update`, and `update` might switch to bisection in the U3 step (as named in the initial HZ publication). If we did bisect, the line-search "are we making enough progress?" check (step L2) is nonsensical (we might have shrunk multiple iterations of bisection, but that is not an indication that the secant model was "working").

Consequently, this reports back about whether bisection was engaged in `update`, and if so skip any kind of convergence assessment and do another iteration.

Fixes #173. Sadly there isn't a very portable test.

### DIFF
--- a/src/hagerzhang.jl
+++ b/src/hagerzhang.jl
@@ -203,10 +203,10 @@ function (ls::HagerZhang)(ϕ, ϕdϕ,
         else
             # We'll still going downhill, expand the interval and try again.
             # Reaching this branch means that dphi_c < 0 and phi_c <= phi_0 + ϵ_k
-            # So cold = c has a lower objective than phi_0 up to epsilon. 
+            # So cold = c has a lower objective than phi_0 up to epsilon.
             # This makes it a viable step to return if bracketing fails.
 
-            # Bracketing can fail if no cold < c <= alphamax can be found with finite phi_c and dphi_c. 
+            # Bracketing can fail if no cold < c <= alphamax can be found with finite phi_c and dphi_c.
             # Going back to the loop with c = cold will only result in infinite cycling.
             # So returning (cold, phi_cold) and exiting the line search is the best move.
             cold = c
@@ -266,7 +266,7 @@ function (ls::HagerZhang)(ϕ, ϕdϕ,
             mayterminate[] = false # reset in case another initial guess is used next
             return a, values[ia] # lsr.value[ia]
         end
-        iswolfe, iA, iB = secant2!(ϕdϕ, alphas, values, slopes, ia, ib, phi_lim, delta, sigma, display)
+        iswolfe, iA, iB, bisected = secant2!(ϕdϕ, alphas, values, slopes, ia, ib, phi_lim, delta, sigma, display)
         if iswolfe
             mayterminate[] = false # reset in case another initial guess is used next
             return alphas[iA], values[iA] # lsr.value[iA]
@@ -274,34 +274,39 @@ function (ls::HagerZhang)(ϕ, ϕdϕ,
         A = alphas[iA]
         B = alphas[iB]
         @assert B > A
-        if B - A < gamma * (b - a)
-            if display & LINESEARCH > 0
-                println("Linesearch: secant succeeded")
-            end
-            if nextfloat(values[ia]) >= values[ib] && nextfloat(values[iA]) >= values[iB]
-                # It's so flat, secant didn't do anything useful, time to quit
+        if !bisected
+            if B - A < gamma * (b - a)
                 if display & LINESEARCH > 0
-                    println("Linesearch: secant suggests it's flat")
+                    println("Linesearch: secant succeeded")
                 end
-                mayterminate[] = false # reset in case another initial guess is used next
-                return A, values[iA]
+                if nextfloat(values[ia]) >= values[ib] && nextfloat(values[iA]) >= values[iB]
+                    # It's so flat, secant didn't do anything useful, time to quit
+                    if display & LINESEARCH > 0
+                        println("Linesearch: secant suggests it's flat")
+                    end
+                    mayterminate[] = false # reset in case another initial guess is used next
+                    return A, values[iA]
+                end
+                ia = iA
+                ib = iB
+            else
+                # Secant is converging too slowly, use bisection
+                if display & LINESEARCH > 0
+                    println("Linesearch: secant failed, using bisection")
+                end
+                c = (A + B) / convert(T, 2)
+
+                phi_c, dphi_c = ϕdϕ(c)
+                @assert isfinite(phi_c) && isfinite(dphi_c)
+                push!(alphas, c)
+                push!(values, phi_c)
+                push!(slopes, dphi_c)
+
+                ia, ib = update!(ϕdϕ, alphas, values, slopes, iA, iB, length(alphas), phi_lim, display)
             end
+        else
             ia = iA
             ib = iB
-        else
-            # Secant is converging too slowly, use bisection
-            if display & LINESEARCH > 0
-                println("Linesearch: secant failed, using bisection")
-            end
-            c = (A + B) / convert(T, 2)
-
-            phi_c, dphi_c = ϕdϕ(c)
-            @assert isfinite(phi_c) && isfinite(dphi_c)
-            push!(alphas, c)
-            push!(values, phi_c)
-            push!(slopes, dphi_c)
-
-            ia, ib = update!(ϕdϕ, alphas, values, slopes, iA, iB, length(alphas), phi_lim, display)
         end
         iter += 1
     end
@@ -373,14 +378,15 @@ function secant2!(ϕdϕ,
     push!(slopes, dphi_c)
 
     ic = length(alphas)
+    bisected = false
     if satisfies_wolfe(c, phi_c, dphi_c, phi_0, dphi_0, phi_lim, delta, sigma)
         if display & SECANT2 > 0
             println("secant2: first c satisfied Wolfe conditions")
         end
-        return true, ic, ic
+        return true, ic, ic, bisected
     end
 
-    iA, iB = update!(ϕdϕ, alphas, values, slopes, ia, ib, ic, phi_lim, display)
+    iA, iB, bisected = update!(ϕdϕ, alphas, values, slopes, ia, ib, ic, phi_lim, display)
     if display & SECANT2 > 0
         println("secant2: iA = ", iA, ", iB = ", iB, ", ic = ", ic)
     end
@@ -412,14 +418,14 @@ function secant2!(ϕdϕ,
             if display & SECANT2 > 0
                 println("secant2: second c satisfied Wolfe conditions")
             end
-            return true, ic, ic
+            return true, ic, ic, bisected
         end
         iA, iB = update!(ϕdϕ, alphas, values, slopes, iA, iB, ic, phi_lim, display)
     end
     if display & SECANT2 > 0
         println("secant2 output: a = ", alphas[iA], ", b = ", alphas[iB])
     end
-    return false, iA, iB
+    return false, iA, iB, bisected
 end
 
 # HZ, stages U0-U3
@@ -457,10 +463,10 @@ function update!(ϕdϕ,
                 ", dphi_c = ", dphi_c)
     end
     if c < a || c > b
-        return ia, ib #, 0, 0  # it's out of the bracketing interval
+        return ia, ib, false #, 0, 0  # it's out of the bracketing interval
     end
     if dphi_c >= zeroT
-        return ia, ic #, 0, 0  # replace b with a closer point
+        return ia, ic, false #, 0, 0  # replace b with a closer point
     end
     # We know dphi_c < 0. However, phi may not be monotonic between a
     # and c, so check that the value is also smaller than phi_0.  (It's
@@ -468,11 +474,11 @@ function update!(ϕdϕ,
     # secure environment of alpha=0; that's why we didn't check this
     # above.)
     if phi_c <= phi_lim
-        return ic, ib#, 0, 0  # replace a
+        return ic, ib, false#, 0, 0  # replace a
     end
     # phi_c is bigger than phi_0, which implies that the minimum
     # lies between a and c. Find it via bisection.
-    return bisect!(ϕdϕ, alphas, values, slopes, ia, ic, phi_lim, display)
+    return (bisect!(ϕdϕ, alphas, values, slopes, ia, ic, phi_lim, display)..., true)
 end
 
 # HZ, stage U3 (with theta=0.5)


pkofod:
Thanks @timholy really appreciate it. I'll try to find time to review it later today .

mateuszbaran:
Thank you! This really turned out to be more involved than I initially expected.

mateuszbaran:
Bump?

pkofod:
Okay, this looks good to me. @mateuszbaran did you somehow check if this solved your original problem?

mateuszbaran:
I've just checked. We've solved the original problem by making L-BFGS in Manopt more robust to line searches returning stepsize 0, so it's harder to reproduce the true original problem, but I can still generate problems for which the sub-optimal alpha is selected in the "It's so flat, secant didn't do anything useful, time to quit" branch.

mateuszbaran:
This script can easily generate cases where such choice is made:
```julia
using Revise
using Manopt
using Manifolds
using LineSearches
using LinearAlgebra

norm_inf(M::AbstractManifold, p, X) = norm(X, Inf)

function f_rosenbrock(x)
    result = 0.0
    for i in 1:2:length(x)
        result += (1.0 - x[i])^2 + 100.0 * (x[i + 1] - x[i]^2)^2
    end
    return result
end
function f_rosenbrock(::AbstractManifold, x)
    return f_rosenbrock(x)
end

function g_rosenbrock!(storage, x)
    for i in 1:2:length(x)
        storage[i] = -2.0 * (1.0 - x[i]) - 400.0 * (x[i + 1] - x[i]^2) * x[i]
        storage[i + 1] = 200.0 * (x[i + 1] - x[i]^2)
    end
    return storage
end

function g_rosenbrock!(M::AbstractManifold, storage, x)
    g_rosenbrock!(storage, x)
    riemannian_gradient!(M, storage, x, storage)
    return storage
end

function test_case_manopt()
    for i in 2:2:1000
        N = 5000+i
        mem_len = rand(4:10)
        println("i = $i, mem_len=$mem_len")
        M = Manifolds.Sphere(N - 1; parameter=:field)
        ls_hz = LineSearches.HagerZhang()

        x0 = zeros(N)
        x0[1] = 1
        manopt_sc = StopWhenGradientNormLess(1e-6; norm=norm_inf) | StopAfterIteration(1000)

        quasi_Newton(
            M,
            f_rosenbrock,
            g_rosenbrock!,
            x0;
            stepsize=Manopt.LineSearchesStepsize(ls_hz),
            #stepsize=HagerZhangLinesearch(M),
            evaluation=InplaceEvaluation(),
            vector_transport_method=ParallelTransport(),
            return_state=true,
            memory_size=mem_len,
            stopping_criterion=manopt_sc,
        )
    end
end
test_case_manopt()
```
You can then add `if values[ib] < values[iA] || values[iB] < values[iA]` check in HZ to see if the choice is indeed suboptimal.

pkofod:
> I've just checked. We've solved the original problem by making L-BFGS in Manopt more robust to line searches returning stepsize 0, so it's harder to reproduce the true original problem, but I can still generate problems for which the sub-optimal alpha is selected in the "It's so flat, secant didn't do anything useful, time to quit" branch.

have you decided to reset the approximation in that case?

mateuszbaran:
I had a chat about it with Roland Herzog some time ago and he suggested using truncated CG is such case but restarting is also a valid option.

pkofod:
@timholy do you have any thoughts on the above comments by @mateuszbaran ?

mateuszbaran:
Bump?

pkofod:
> I've checked the paper and it doesn't have the termination condition corresponding to "It's so flat, secant didn't do anything useful, time to quit" present in LineSearches.jl implementation. I don't see what the justification for it is. Maybe it should be modified?

I think Tim may have added that. I suppose you're saying that maybe it's failing here partially because the branch doesn't belong?

mateuszbaran:
There is clearly a difference between how it's commented and what it actually does. It's possible that termination condition actually catches some corner and in some cases is a good idea but currently it also catches things it shouldn't catch (including in this PR). I don't know that corner case it was supposed to handle so I don't know how to properly fix it. 

Anyway, there is always that simple change I've done that works well enough for me and shouldn't really harm any use case.

There is already at least one person (other than me and Ronny) who had issues because my fixed HZ linesearch is what works best for his problem but he would like to use it in Pluto which doesn't like non-registered packages. So it would be nice to have a good HZ linesearch in a registered package.

timholy:
Sorry I haven't had time for "general"/"charitable" development lately. My busy schedule is likely to persist another couple of weeks at least. If anyone wants to take this effort over, by all means go for it.

A couple of high-level thoughts:
- the Julia implementaton of HZ linesearch should be in this package, and it should be good. It's important to fix this and release a new version.
- this episode is a reminder that in some ways, the test suite is more valuable than the implementation. My extensions to the published algorithm were almost certainly motivated by real-world cases, but those are lost to the mists of time. My inability to reconstruct those motivations is at risk of paralyzing any decision here. If push comes to shove, it might be better to revert to the published algorithm and force me to deal with the consequences.
- Likewise, the issues uncovered by @mateuszbaran depend on a complicated software stack that may be more than one wants to depend on for a test suite: updates to a large ecosystem can have a way of making such things fragile. So we're at risk for losing yet more important test cases if we don't do something.
- To me, it seems that the most important step is to develop a testing layer where we capture the outcome of the one-dimensional problem, as that removes dependence from the "real" problem and allows us to create small tests without a large dependency stack. The idea would be to grab the record of the `phi` and `dphi` values at the `alpha` values chosen by the algorithm, and use interpolation to match both the values and slopes at those `alpha` values. (IIRC Lagrange polynomial interpolation can be readily generalized to match both values and derivatives.) Then we could have a test suite that consists of vectors of `alpha`, `phi`, and `dphi` values for all the "bad" cases anyone reports.
- Failing to do this now, while we have real-world failing test cases, is just a way of perpetuating the inability to keep this package at the forefront of Julia implementations of line search methods. Which would be a giant waste.

Thus if someone wants to help move this forward, perhaps the best thing to do is focus on developing that test harness.

mateuszbaran:
Thank you, your observations are very on-point. To push this forward I can prepare a simplified test case, either this week or the next one. Anyway, I'm not a maintainer of this package so I can't decide how to fix this issue.

mateuszbaran:
I've updated my example, now there is a clear, fully self-contained case of sub-optimal returned value.

pkofod:
Thank you so much Mateusz! I hope to have a little more time next week than in the past months :)

pkofod:
Not forgotten! I will use your test-case to look at this next week.

pkofod:
> I've updated my example, now there is a clear, fully self-contained case of sub-optimal returned value.

You have been patient :) I had some deadlines that were quite important. I fixed some Optim.jl stuff yesterday evening, and I will try to look at this when I'm off work.

kbarros:
As per https://github.com/JuliaNLSolvers/LineSearches.jl/issues/175, I encountered another issue where the flatness check causes early termination at a point that is not close to stationary. However, this PR does _not_ seem to resolve that other flatness check issue.

The problematic sequence of `ϕ` and `dϕ` data I found are these:

```jl
cs = [0, 0.2, 0.1, 0.055223623837016025]
ϕs = [3.042968312396456, 3.117411287254072, -3.503584823341612, 0.5244246783084083]
dϕs = [-832.4270136930788, -505.33622492291033, 674.947830358913, 738.3388472434362]
```

The original function is oscillatory, but like @timholy says above, one can just as well create a polynomial giving the same data.

> The idea would be to grab the record of the phi and dphi values at the alpha values chosen by the algorithm, and use interpolation to match both the values and slopes at those alpha values. (IIRC Lagrange polynomial interpolation can be readily generalized to match both values and derivatives.)

I created the package [HermiteInterpolation](https://github.com/kbarros/HermiteInterpolation.jl) for exactly this purpose (submitted for registration), and can be used like this:

```jl
using HermiteInterpolation: fit
f = fit(cs, ϕs, dϕs)
@assert f.(cs) ≈ ϕs

using DynamicPolynomials
@polyvar c

println(f(c))
# 3.042968312396456 - 832.4270136930788*c - 132807.15591801773*c^2 + 7.915421661743959e6*c^3 - 1.570284840040962e8*c^4 + 1.4221708747294645e9*c^5 - 5.970065591205604e9*c^6 + 9.405512899903618e9*c^7

println(differentiate(f(c), c))
# -832.4270136930788 - 265614.31183603546*c + 2.3746264985231876e7*c^2 - 6.281139360163848e8*c^3 + 7.110854373647323e9*c^4 - 3.582039354723362e10*c^5 + 6.5838590299325325e10*c^6
```

In the wild, this type of bug was extremely time-consuming to diagnose -- the last thing I was suspecting was a problem in LineSearches. I was expecting instead a problem with the way I did the memory mutation in gradient evaluation code. Maybe as a band-aid we can add a check inside that flatness check -- if the norm of the gradient does not seem small, then print a warning for the user that there is a known bug with this part of LineSearches?

kbarros:
Here is an optimization problem that highlights a fundamental problem with the current flatness check:

```jl
# The minimizer is x0=[0, 0], with f(x0) = 1
function f(x)
    return (x[1]^2 + 1) * (2 - cos(100*x[2]))
end

using Optim

function test_converges(method)
    for i in 1:100
        r = randn(2)
        res = optimize(f, r, method)
        if Optim.converged(res) && minimum(res) > f([0,0])
            println("""
                Incorrectly reported convergence after $(res.iterations) iterations
                Reached x = $r with f(x) = $(minimum(res))
                """)
        end
    end
end

# Works successfully, no printed output
test_converges(LBFGS(; linesearch=Optim.LineSearches.BackTracking(order=2)))

# Prints ~10 failures to converge (in 100 tries). Frequently fails after the
# first line search.
test_converges(ConjugateGradient())
```

timholy:
I'll get to this in another week or two.

pkofod:
> Here is an optimization problem that highlights a fundamental problem with the current flatness check:

I think this may highlight the same issue as in https://github.com/JuliaNLSolvers/LineSearches.jl/pull/174#issuecomment-2043311432

> Here is an optimization problem that highlights a fundamental problem with the current flatness check:

With my own simple branch made for checking this I am able to run

```
julia> test_converges(ConjugateGradient(; linesearch=Optim.LineSearches.HagerZhang(check_flatness=false)))

julia> test_converges(ConjugateGradient(; linesearch=Optim.LineSearches.HagerZhang(check_flatness=true)))
Incorrectly reported convergence after 1 iterations
Reached x = [0.16013274827601717, 0.35097409719940265] with f(x) = 2.9310451880340986

Incorrectly reported convergence after 1 iterations
Reached x = [0.0507085888328239, 0.1084483256815114] with f(x) = 2.1557003490556643

Incorrectly reported convergence after 2 iterations
Reached x = [0.7211228462970422, -22.3471576683635] with f(x) = 3.8050518641095663

Incorrectly reported convergence after 1 iterations
Reached x = [-1.2589697390692955, -1.0403076610453794] with f(x) = 7.5909348717323635

Incorrectly reported convergence after 1 iterations
Reached x = [0.08409600277401064, 0.10725179410260108] with f(x) = 2.2831453152517023

Incorrectly reported convergence after 2 iterations
Reached x = [0.9010705241444945, 0.7330566636590027] with f(x) = 4.526934892232607

Incorrectly reported convergence after 2 iterations
Reached x = [-0.27240989049521974, 11.156588470976192] with f(x) = 3.1405412374583666

Incorrectly reported convergence after 2 iterations
Reached x = [0.4717492439353791, -15.742921266415511] with f(x) = 3.5917484672484252

Incorrectly reported convergence after 1 iterations
Reached x = [-1.4143709712675014, 0.3441070259740454] with f(x) = 8.969056311635578

Incorrectly reported convergence after 1 iterations
Reached x = [0.29458492668827274, 1.1477725568576247] with f(x) = 2.291918955272849

Incorrectly reported convergence after 1 iterations
Reached x = [-0.19214192623076573, -0.14209344321079873] with f(x) = 2.1486141359671405

Incorrectly reported convergence after 4 iterations
Reached x = [-0.21665091042117182, 43.27319699842477] with f(x) = 2.3266172361393616

Incorrectly reported convergence after 1 iterations
Reached x = [-0.6789087745560297, 0.5133014956972761] with f(x) = 2.213588359372836
```

and running @mateuszbaran problem here https://github.com/JuliaNLSolvers/LineSearches.jl/pull/174#issuecomment-2043311432 I get

```
julia> res, res_cache = test_tc(tc1, true)
0.0 3003.592409634743 -22332.321416890798
1.0 2962.0378569864743 -20423.214551925797
5.0 2891.4462095232184 11718.185026267562
3.541670844449739 3000.9760725116876 -22286.821227217057
((3.541670844449739, 3000.9760725116876), LineSearchCache{Float64}([0.0, 1.0, 5.0, 3.541670844449739], [3003.592409634743, 2962.0378569864743, 2891.4462095232184, 3000.9760725116876], [-22332.321416890798, -20423.214551925797, 11718.185026267562, -22286.821227217057]))

julia> @test_broken minimum(res_cache.values) == res[2]

Test Broken
  Expression: minimum(res_cache.values) == res[2]

julia> res2, res_cache2 = test_tc(tc1, false)
0.0 3003.592409634743 -22332.321416890798
1.0 2962.0378569864743 -20423.214551925797
5.0 2891.4462095232184 11718.185026267562
3.541670844449739 3000.9760725116876 -22286.821227217057
4.49745720537289 -2137.605778336924 7059.04612357025
((4.49745720537289, -2137.605778336924), LineSearchCache{Float64}([0.0, 1.0, 5.0, 3.541670844449739, 4.49745720537289], [3003.592409634743, 2962.0378569864743, 2891.4462095232184, 3000.9760725116876, -2137.605778336924], [-22332.321416890798, -20423.214551925797, 11718.185026267562, -22286.821227217057, 7059.04612357025]))

julia> @test minimum(res_cache2.values) == res2[2]
Test Passed
```

with this plot of the problematic case

image

so based on the plot and the order of alphas above we see that we first check 1 that has downwards slope then 5 that has upwards. We additionally seem to come up with 3.54... that has downwards slop and then we get a bracket [3.54..., 5.0] that must contain a minimum. But nothing here suggests that the function should be flat between 3.54... and 5.0. The flatness check seems to check an unfortunate condition that

```
nextfloat(3000.9760725116876) >=  2891.4462095232184
```

in other words A is 3.54... and B = 5.0 so B > A as suggested in the assertion but why would we know that 
```
nextfloat(values[ia]) >= values[ib] && nextfloat(values[iA]) >= values[iB]
```
is then an appropriate test for flatness if `nextfloat(values[iA]) >= values[iB]`? It seems to expect that `values[iA] == values[iB]` or `values[iA] < values[iB]` or am I misunderstanding something in this flatness check?

timholy:
Yes, I think it's making some assumptions about what those indices mean.

pkofod:
I think (ia, ib, iA, iB) = (0, 1, 3.54..., 5.0) above but I have to go back and print :)

kbarros:
The experiment in #182 suggests that simply disabling the flatness check would be an improvement in behavior. Is there a blocker for merging?  Perhaps some tests can be included from #180.

pkofod:
> The experiment in #182 suggests that simply disabling the flatness check would be an improvement in behavior. Is there a blocker for merging? Perhaps some tests can be included from #180.

I was a bit busy for the past month. I will apply the change with an escape hatch to get the old behavior and we can improve from there.

pkofod:
> Hi @pkofod, great news that you're pushing this forward! This change may break certain tests in [Sunny.jl](https://github.com/SunnySuite/Sunny.jl). If you're not doing a major version bump of this package and Optim.jl, would you mind coordinating here on the package releases so that I can ensure a smooth transition? Another possibility is to make the escape hatch go the other direction -- have the changes be "opt-in" for some transition period?

We could introduce it as a "non-change" but the change is made to fix a "bug" or "buggy" behavior so technically I think it should be fine. The current behavior seems to have caused more bad than good so I am comfortable doing it in a feature release as a fix + feature to return to the old behavior. Semver tells you that you need to tag a breaking change if the user needs to change their code in the new version but not if it's because a bug was fixed. 

I can tag it first with the feature but without changing the default and then later change it, but it shouldn't really matter I think. You should be able to have bounds in your Project to avoid getting the new version.

Can you open an issue on your repo to describe what the issue is exactly? I will be happy to wait with the tag until we can resolve this.

pkofod:
I think this has been resolved with the hz cleanup

mateuszbaran:
Oh, you also have recently reimplemented HZ? I did it recently in https://github.com/JuliaManifolds/Manopt.jl/pull/554 because the issues with LS were bothering me. It eventually was performing better than the old HZ from LS. See here for some changes I've made compared to the original paper: https://github.com/JuliaManifolds/Manopt.jl/pull/554/changes#diff-53c4195b90bd55528c4074da698f8df678db449b02eb15b1cb9ad239420605ffR2752-R2773 .

BTW, we will have L-BFGS-B in Manopt 0.6 :slightly_smiling_face: . It's not tagged yet because we are planning a couple more breaking changes but it performs quite well: https://arxiv.org/abs/2605.10573 .

pkofod:
> BTW, we will have L-BFGS-B in Manopt 0.6 🙂 . It's not tagged yet because we are planning a couple more breaking changes but it performs quite well: https://arxiv.org/abs/2605.10573 .

Cool! I have been looking into the same :) 

> See here for some changes I've made compared to the original paper: https://github.com/JuliaManifolds/Manopt.jl/pull/554/changes#diff-53c4195b90bd55528c4074da698f8df678db449b02eb15b1cb9ad239420605ffR2752-R2773 .

Indeed, published algorithms need to be presented in some form but are not necessarily perfect :) 

In the end, I ended up re-implementing it to sanely arrive at a conclusion. The two papers are actually very well structured, so it was not too bad in the end :) I'm sure you'll enjoy full control in your own repository! Thanks for helping diagnosing it anyway.

PR #177, opened 2024-07-29
Add LineSearchTestCase

### LINKED ISSUE #102: Add logging functionality
I'd like to see a flexible logging / tracing functionality here. 

Currently there is little info provided when things go wrong inside the linesearches. Finiteness tests such as #101 should (optionally) warn the user that something is wrong.

timholy:
Also includes the failing case in PR#174.

This is minimally modified from the code provided by @mateuszbaran (who I've listed as a co-author).

~~One thing that seems missing: documentation on how to capture the specific values tried by the failing line search method. I'm not broadly familiar with them all, is there any good guidance on that? Or should we add a `cache::Bool` option to them all?~~

Once this gets merged, we should first capture any other problem reports and then start trying to fix the failures.

Closes #102

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 docs/build
 docs/src/examples/generated
 /docs/Manifest.toml
+Manifest.toml
--- a/Project.toml
+++ b/Project.toml
@@ -13,6 +13,8 @@ Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 DoubleFloats = "1"
 NLSolversBase = "7"
 NaNMath = "1"
+Optim = "1"
+OptimTestProblems = "2"
 Parameters = "0.10, 0.11, 0.12"
 julia = "1.6"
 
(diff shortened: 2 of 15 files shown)


mateuszbaran:
Thanks for continuing your work on this issue :+1: . I would generally hope that line search failures would be infrequent enough that convenience options for saving values used for interpolation wouldn't be necessary? After this PR it would be enough to `]dev LineSearches`, add `println` instructions to the line search and copy the three vectors to `LineSearchTestCase`. It's still much less work than figuring out that the line search might be the issue, so probably documenting this workflow would be sufficient?

timholy:
I've also found a weird NaN bug with `MoreThuente` (might be in my code, might not), so it may be a broader issue? And `MoreThuente` doesn't seem to cache the values so it may be less straightforward to capture.

timholy:
OK, I think this is ready to merge. I don't have privileges myself so I'll rely on @pkofod to review and make final decisions.

pkofod:
Thanks! I'll review

PR #180, opened 2024-10-22
Add test cases, disable flatness check in HZ

timholy:
This may cause problems of its own, but for the time being it's better
than the status quo.

Fixes #173
Closes #174
Fixes #175

I am pretty sure that without the flatness check, I've seen HZ iterate until the `linesearchmax` limit is hit, which is rather inefficient. This comes nowhere close to hitting that limit, though. So until we have a better test case we should probably disable the flatness check.

Note: this PR should probably not be squash-merged, the three commits are crafted to be independent.

PR #182, opened 2025-03-26
Add check_flatness keyword to HZ and format

pkofod:
See https://github.com/JuliaNLSolvers/LineSearches.jl/pull/174#issuecomment-2753323452

I'm not sure the check does what it's intended to do, so I'm turning it off and giving the possibility of turning it back on.


pkofod:
I should have not done the format in this pr, but... I just add a keyword and a test. That said, of course the tests do not pass on CI 😵

pkofod:
> I should have not done the format in this pr, but... I just add a keyword and a test. That said, of course the tests do not pass on CI 😵

i see, it also fails when in a test set locally but not when run by hand 👍 what a gift that keeps on giving :)

pkofod:
> > I should have not done the format in this pr, but... I just add a keyword and a test. That said, of course the tests do not pass on CI 😵
> 
> i see, it also fails when in a test set locally but not when run by hand 👍 what a gift that keeps on giving :)

nevermind, just seems like issues.jl was not loaded and the same test had already been added as broken somewhere else

pkofod:
This is just a quick'n'dirty "solution". #180 is a re-implementation as mine in NLSolvers.jl and I want to compare them later, but I don't think the flatness check is currently correct so I think we should turn it off by default

PR #187, opened 2025-11-13
Calculate derivatives using Jacobian-vector products

devmotion:
A sketch of how https://github.com/JuliaNLSolvers/NLSolversBase.jl/pull/168 could be used in LineSearches.jl.

A possible downside could be that `gradient(df)` would not return the gradient of the objective function at `x_new` anymore, so would be computed in e.g. Optim.jl after the line search is completed.


pkofod:
So, for example, I don't think this should be changed https://github.com/JuliaNLSolvers/Optim.jl/blob/b367149cfec6e339ca3790714d1625d82a8b3f69/src/utilities/perform_linesearch.jl#L42 that relies og g(x) being readily available (free/lookup) because we already evaluated it in the past iteration / first iteration to check for convergence.

pkofod:
We have benchmarks in Optim, but not here. Would be interesting to see the effect, but on the other hand maybe everything is too fast.

pkofod:
This is ready for review I suppose?

devmotion:
No, NLSolversBase is not ready yet.

pkofod:
looks good to go but is conditional on nlsolversbase being merged and tagged

pkofod (approved):

PR #188, opened 2025-11-22
Require Julia 1.10 and remove Parameters

devmotion:
`Parameters.@unpack` is not needed on Julia >= 1.7 as one can just use the native `(; ...) = ...` syntax; similarly one can use `@kwdef` instead of `Parameters.@with_kw`.

In fact, `@unpack` (which was factored out to `UnPack` at some point) creates `Val` dispatches for the property accesses which at some point caused compilation time issues in OrdinaryDiffEq; since that was prior to Julia 1.7, we replaced it with `SimpleUnPack.@unpack` for a while: https://github.com/devmotion/SimpleUnPack.jl

Currently, the package supports Julia 1.6, so to be able to use `(; ...) = ...` and `@kwdef` I updated the Julia compat entry to 1.10 and also updated CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,24 +2,24 @@ name: Run tests
 
 on: [push, pull_request]
 
+# needed to allow julia-actions/cache to delete old caches that it has created
+permissions:
+  actions: write
+  contents: read
+
 jobs:
   test:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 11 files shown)


devmotion:
@pkofod I haven't contributed to this repo yet, could you trigger the CI workflow in this PR?

pkofod:
#189 has these changes already

PR #190, opened 2025-11-26
Remove use of implicit `NLSolversBase.value` and `NLSolversBase.gradient`

devmotion:
Extracted from #187

### DIFF
--- a/src/initialguess.jl
+++ b/src/initialguess.jl
@@ -74,7 +74,7 @@ function (is::InitialQuadratic{T})(ls, state, phi_0, dphi_0, df) where T
         # If we're at the first iteration
         αguess = is.α0
     else
-        αguess = 2 * (NLSolversBase.value(df) - state.f_x_previous) / dphi_0
+        αguess = 2 * (phi_0 - state.f_x_previous) / dphi_0
         αguess = NaNMath.max(is.αmin, state.alpha*is.ρ, αguess)
         αguess = NaNMath.min(is.αmax, αguess)
         # if αguess ≈ 1, then make it 1 (Newton-type behaviour)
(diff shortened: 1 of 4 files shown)


src/initialguess.jl
```diff
@@ -74,7 +74,8 @@ function (is::InitialQuadratic{T})(ls, state, phi_0, dphi_0, df) where T
         # If we're at the first iteration
         αguess = is.α0
     else
-        αguess = 2 * (NLSolversBase.value(df) - state.f_x_previous) / dphi_0
+        f_x_ls = NLSolversBase.value!(df, state.x_ls)
```
devmotion: @pkofod Is `state.x_ls` the point that should be evaluated here?

test/arbitrary_precision.jl
```diff
@@ -98,6 +98,7 @@ doublefloatstypes = [Double64, Double32, Double16]
     ls = HagerZhang{T}()
     state = getstate()
     state.f_x_previous = 2*phi_0
+    state.x_ls = zero(x)
```
devmotion: All these modified tests trigger the branch with `f_x_ls` above. However, that produces incorrect and stochastic results, and errors for `BigFloat`s, unless `x_ls` is initialised with a fixed value.

src/initialguess.jl
pkofod: This should be `state.x`. That is, the x that produced phi0 and dphi0

src/initialguess.jl
```diff
@@ -180,8 +180,10 @@ function (is::InitialHagerZhang)(ls::Tls, state, phi_0, dphi_0, df) where Tls
         # and the user has not provided an initial step size (is.α0 is NaN),
         # then we
         # pick the initial step size according to HZ #I0
-        state.alpha = _hzI0(state.x, NLSolversBase.gradient(df),
-                            NLSolversBase.value(df),
+        # phi_0 is (or should be) equal to NLSolversBase.value(df, state.x) 
+        # TODO: Make the gradient available as part of the state?
```
pkofod, line 184: yes, generally only g_previous is available with the logic being that we can use the cache in the objective, but as we have discussed it makes sense to add an additional vector/array to the algorithm state because it's so much easier to ensure correctness that way. One more array won't mean anything in the full scheme

pkofod (approved):

PR #193, opened 2026-03-17
A re-implemented HZ

pkofod:
@timholy I noticed that in the existing implementation initially the Wolfe conditions are only checked if mayterminate[] when I plotted out solutions of the existing implementation. This seems strange to me. This forces a bracketing step even if the initial state satisfies the conditions we seek to meet. I changed it, but please argue if you think it's wrong.

They should largely be the same, but I'm seeing some differences on some test problems that I'm curious to learn more about.  Once thing I noticed already was the thing mentioned above, but they return different solutions quite often, so something must be different. If there's a bug in either implementation I'm not sure.

Optim tests ran fine with the implementation as well.

I did not implement the initial step size logic (yet)


src/hagerzhang.jl
```diff
@@ -163,12 +163,13 @@ function (ls::HagerZhang)(ϕ, ϕdϕ,
 
     # If c was generated by quadratic interpolation, check whether it
     # satisfies the Wolfe conditions
-    if mayterminate[] &&
-          satisfies_wolfe(c, phi_c, dphi_c, phi_0, dphi_0, phi_lim, delta, sigma)
+    if satisfies_wolfe(c, phi_c, dphi_c, phi_0, dphi_0, phi_lim, delta, sigma)
```
timholy, line 173: Is it possible that the initial value of `mayterminate` should be changed to `true`? It's been ages so I can't remember why it starts the way it does.

pkofod: I was actually wondering if it should, yes, but there are many branches that set it to false. But maybe if you're using InitialStatic for example (so you're not using the quadric stuff from HZ the original article about conjugate gradient descent, then true is maybe appropriate on all entries.

pkofod: The interval increment is very large by default. That is also a matter of debate of course. Maybe appropriate for CGD but maybe not for quasi-Newton where 1 may be reasonable once you have accumulated some curvature information, but 5 may be a long step in *that* context.

pkofod: Defaults are of course always a hot topic :)

pkofod:
If HZ falls back to bisection there's a phidphi call. In the new code convergence is checked every time a finite phidphi evaluation occurs. In the old code, only calls inside of secant2 were checked for convergence. Before the change here for the initial check, this meant that HZ *had* to run a secant iteration to converge.

pkofod:
When comparing the existing implementation to the re-implementation I realized that once you actually make sure to return when convergence requirements are met instead of allowing iterations without convergence checking the flatness test fails at that may actually make quite some sense. U3 can be fully iterated through if you're in a flat/noisy region where it keeps on thinking it has passed a hill top. Checking convergence more often makes the test that previously was marked as broken actually pass and matches my reimplementation.

src/LineSearches.jl
```diff
@@ -7,7 +7,7 @@ using NLSolversBase: NLSolversBase, AbstractObjective
 
 export LineSearchException, LineSearchCache
 
-export AbstractLineSearch, BackTracking, HagerZhang, Static, MoreThuente, StrongWolfe
+export AbstractLineSearch, BackTracking, HagerZhang, HagerZhangLS, Static, MoreThuente, StrongWolfe
```
pkofod:
```suggestion
export AbstractLineSearch, BackTracking, HagerZhang, Static, MoreThuente, StrongWolfe
```

src/hagerzhang.jl
```diff
@@ -163,12 +170,13 @@ function (ls::HagerZhang)(ϕ, ϕdϕ,
...
```
JoshuaLampert, line 173: We are seeing failing tests in https://github.com/ranocha/SummationByPartsOperators.jl/pull/414, where we use Optim.jl with `LBFGS` and default line search algorithm `HagerZhang()`. With LineSearches.jl v7.6.2 the optimization needs almost three times as many iterations to properly converge (4309 vs 12083 iterations) in one of our examples. I verified locally that this change (i.e., putting the `mayterminate[]` outside of the if condition) is responsible for that. I do not quite understand all details, but can something be done against this? I also tried to use `LBFGS(linesearch=HagerZhang(mayterminate=Ref(true))` instead, but this did not change anything in the result.

JoshuaLampert: Note that we are not the only ones with failing CI due to v7.6.2 (probably also because of this PR and presumably also because of this exact line), see https://github.com/aws-cqc/DeviceLayout.jl/pull/217 and https://github.com/JuliaNLSolvers/Optim.jl/pull/1253. So can this change maybe be reverted or otherwise fixed, @pkofod?

pkofod: I can have a look, but it is of course always hard to diagnose based on objectives I don't fully understand

pkofod: can you explain what you did to make it "work" ? I want to understand what you mean, because I don't think reverting this change is correct. What happened previously was that hagerzhang failed to check initial convergence. Yes, this meant that you often found a more accurate solution per line search, but that was not intended as such.

pkofod: Let me put it in a different way. If you in optimize set g_tol = 1-2 but internally we ignore the check for the first 100 iterations, then it is indeed true that many invocations will end up at a more accurate solution (if g_tol was met prior to the 100th iteration), but it is of course not the correct behavior of the code.

JoshuaLampert: > I can have a look, but it is of course always hard to diagnose based on objectives I don't fully understand

Yes, I fully understand that, but it can also be a bit hard to construct simple-to-understand objective functions, which exhibit the same/similar behavior as the objective function I am interested in.

> can you explain what you did to make it "work" ?

I just applied this diff and for my example, I was back again at the pre-v7.6.2 behavior, where the optimization converged with fewer iterations (and fewer function and gradient evaluations) to a solution I want:
```diff
-    if satisfies_wolfe(c, phi_c, dphi_c, phi_0, dphi_0, phi_lim, delta, sigma)
+    if mayterminate[] &&
+          satisfies_wolfe(c, phi_c, dphi_c, phi_0, dphi_0, phi_lim, delta, sigma)
         if display & LINESEARCH > 0
             println("Wolfe condition satisfied on point alpha = ", c)
         end
-        if mayterminate[] # this previously was in front of satisfies_wolfe but it should be fine to always check initial convergence
-            mayterminate[] = false # reset in case another initial guess is used next
-        end
+        mayterminate[] = false # reset in case another initial guess is used next
         return c, phi_c # phi_c
     end
```
This is probably not a general fix because you obviously did this change deliberately, but at least in my (admittedly, very specific) case I think the previous version behaved superior.

PR #194, opened 2026-03-30
Test convergence more often in HagerZhang to avoid many iterations after a point that satisfies the conditions has already been found

pkofod:


pkofod:
It seems logical to check for convergence whenever ϕdϕ is evaluated as we are not looking for an exact minimizer we're only looking for one that satisfies the wolfe conditions and you can tighten those if you want to. It also seems to make the test_broken for flatness go away which may actually make sense.

PR #202, opened 2026-05-22
Fix sign error and cancellation in BackTracking cubic interpolation

pkofod:
Implements #172 so sorry for not following it all the way to the door sooner @AlexandreMagueresse

The cubic-interpolation branch of BackTracking had two bugs:

1. In the degenerate (a ≈ 0) case, α_tmp = dϕ_0 / (2*b) was missing a minus sign — the minimizer of the resulting quadratic is -ϕ'(0) / (2b), not +ϕ'(0) / (2b).

2. In the general case, the Nocedal-Wright root expression (-b + √(b² − 3aϕ'(0))) / (3a) suffers catastrophic cancellation when ϕ'(0) < 0 and b > 0 (the discriminant exceeds b², so √Δ approaches |b| and the numerator is a difference of close positives). Rewriting via the conjugate as -ϕ'(0)/(b + √Δ) avoids the cancellation and reduces cleanly to the degenerate form as a → 0.

Adds a regression test using the example provided by Alexandre in PR #172 (ϕ(x) = -sin(x)/x, α₀ = 1e5, x₀ = -3): the old code lands at α ≈ 8.43, the fix at α ≈ 2.18, a meaningfully better minimizer.

### DIFF
--- a/src/backtracking.jl
+++ b/src/backtracking.jl
@@ -24,10 +24,10 @@ function (ls::BackTracking)(df::AbstractObjective, x::AbstractArray{T}, s::Abstr
     ϕ, dϕ = make_ϕ_dϕ(df, x_new, x, s)
 
     if ϕ_0 == nothing
-        ϕ_0 = ϕ(Tα(0))
+        ϕ_0 = ϕ(zero(Tα))
     end
     if dϕ_0 == nothing
-        dϕ_0 = dϕ(Tα(0))
+        dϕ_0 = dϕ(zero(Tα))
     end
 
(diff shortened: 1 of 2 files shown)


pkofod:
I also appreciated your cleanup of the constructor / runtime logic but I'm including the bug fix here alone.

AlexandreMagueresse:
Awesome, thanks for merging these fixes!

pkofod:
> Awesome, thanks for merging these fixes!

I appreciate you pointing it out!

PR #213, opened 2026-09-13
HagerZhang: do not evaluate a secant point that is not strictly inside the bracket

andreasnoack:
# HagerZhang: do not evaluate a secant point that is not strictly inside the bracket

Follow-up to #211 / #212.

## The problem

`secant2!` computes the secant root `c` and calls `ϕdϕ(c)` before checking where `c` landed. Only afterwards does `update!` classify it, and since #212 it rejects a `c` that is not strictly interior. So a secant root that rounds onto an endpoint costs a full objective evaluation whose result is thrown away.

The root lands on an endpoint in two ordinary situations: the bracket is a few ulps wide, or one slope is negligible against the other, so `(a * dphi_b - b * dphi_a) / (dphi_b - dphi_a)` rounds to `a` or `b`. The value and slope at both endpoints are already stored, so there is nothing to learn from evaluating there.

This is also the evaluation that exposed #211. With `a == 0`, the wasted call re-evaluates the objective at `x + 0 * s == x`, and if the slope it returns differs from the stored `dphi_0` by rounding, it can differ in sign once the true slope is at the noise floor. #212 made `update!` robust to that; this PR removes the evaluation that produced the disagreement in the first place. With it, nothing in HagerZhang ever re-evaluates the objective at `alpha = 0`, so the route by which the caller computed `dphi_0` cannot matter to the line search.

## The change

- First secant: treat a non-interior `c` (`!(a < c < b)`) like the existing non-finite case and bisect instead. The non-finite check is subsumed, since `a < NaN < b` is false.
- Second secant: require `a < c < b` rather than `a <= c <= b` before evaluating. The aside in #211 already noted this evaluation is discarded by `update!`.

The trajectory of a line search that never produces an endpoint root is unchanged. When one is produced, the bisection midpoint replaces the discarded endpoint evaluation, so the evaluation count does not increase.

## Test

A new testset in `test/errors.jl` builds a bracket `[1.0, nextfloat(1.0, 2)]` with slopes chosen so the secant root rounds exactly onto `a` in one case and onto `b` in the other, checks that it does, then calls `secant2!` with a recording `ϕdϕ` and asserts every evaluation is strictly interior. On master both cases fail (the endpoint is evaluated); on this branch they pass.

## Related

The same measurement that motivated JuliaNLSolvers/Optim.jl#1284 shows the two slope routes differ at the level of the objective's own derivative noise (it tracks the ODE solver tolerance of that objective one for one), so re-evaluating at an endpoint could never have given a more trustworthy slope than the stored one. JuliaNLSolvers/Optim.jl#1286 reverts #1285 on that basis.


src/hagerzhang.jl
```diff
@@ -445,7 +450,8 @@ function secant2!(ϕdϕ,
         # we updated a, do it for b too
         c = secant(alphas, slopes, ia, iA)
     end
-    if (iA == ic || iB == ic) && a <= c <= b
+    # Only an interior second secant point is worth an evaluation, for the same reason
```
pkofod:
```suggestion
    # Only an interior second secant point is worth an evaluation, for the same reason as the first secant call above
```

pkofod (approved):

